PR #101, opened 2015-01-08
Changes for PCM-407

renujhamtani:
@ardasgup @sshumake  Please review and merge.

### DIFF
--- a/app/cases/controllers/commentsSection.js
+++ b/app/cases/controllers/commentsSection.js
@@ -30,6 +30,34 @@ angular.module('RedhatAccess.cases').controller('CommentsSection', [
             });
         };
 
+        $scope.commentReply = function(id,browserIE) {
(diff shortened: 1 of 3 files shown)


app/cases/controllers/commentsSection.js
```diff
@@ -30,6 +30,34 @@ angular.module('RedhatAccess.cases').controller('CommentsSection', [
             });
         };
```
sshumake: Renu this code seems a bit funky in a controller.  From my understanding DOM manipulations should not occur in controllers.  What is the reason this was moved out of the directive?   Also, I know you didnt write this code but how much of a pain would it be to re-write to follow angular best-practices?

sshumake: @renujhamtani awesome, looks good now.  Merging

PR #112, opened 2015-01-15
Fix for PCM-450

ardasgup:
This is a critical defect and should probably be hot-fixed in prod.
@sshumake @renujhamtani  please review and do the needful.


renujhamtani:
Reviewed!!!

PR #113, opened 2015-01-15
403 and 404 errors enabled

sshumake:
@ardasgup @renujhamtani please review

### DIFF
--- a/app/cases/cases.module.js
+++ b/app/cases/cases.module.js
@@ -95,5 +95,13 @@ angular.module('RedhatAccess.cases', [
             controller: 'EditGroup',
             templateUrl: 'cases/views/editGroup.html'
         });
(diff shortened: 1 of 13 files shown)


sshumake:
@ardasgup There was an email thread on pcm-list earlier this week with the subject "pcm-210" addressing these questions.  We cannot link directly to the customer portal errors without decoupling ourselves from in-produc, which we want to discourage.  Also if we use the portal error pages this breaks our "one page app".  A 403 is tied to the manage support case role, can you think of any case where the user would have this role but be getting a 403 for any resource?

renujhamtani:
@sshumake @ardasgup  As aritro on leave, assigning this to Pranjal to review. 

@pbathia Please review this.

PR #115, opened 2015-01-19
Faulty logicc in pcm 210

sshumake:
@renujhamtani @pbathia There is faulty logic in my initial commit.


pbathia:
@sshumake Looks good.

renujhamtani:
@sshumake  Looks good

PR #121, opened 2015-01-23
removing sonar properties file as it is no longer needed.

pbathia:
deleted:    sonar-project.properties


pbathia:
@sshumake please merge.

PR #123, opened 2015-01-27
Fix for PCM-185 and PCM-186

ardasgup:
@sshumake @renujhamtani   please review and merge !

### DIFF
--- a/app/cases/controllers/new.js
+++ b/app/cases/controllers/new.js
@@ -462,7 +462,7 @@ angular.module('RedhatAccess.cases').controller('New', [
             $scope.showRecommendationPanel = true;
         };
 
-        $rootScope.$on(CASE_EVENTS.fetchProductsForContact, function() {
+        $scope.$on(CASE_EVENTS.fetchProductsForContact, function() {
             $scope.productsLoading = true;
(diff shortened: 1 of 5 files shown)


app/cases/controllers/new.js
```diff
@@ -460,5 +461,25 @@ angular.module('RedhatAccess.cases').controller('New', [
         $scope.makeRecommendationPanelVisible =function(){
             $scope.showRecommendationPanel = true;
         };
+
+        $rootScope.$on(CASE_EVENTS.fetchProductsForContact, function() {
```
sshumake: In talking with @connyay I think we have been doing this wrong throughout the code and it has contributed to memory leaks.  @ardasgup please change this to $scope.$on() and see if it works.  If you use $rootscope.$on() you must destroy the listener.  We should investigate this further in our hardening sprint.

ardasgup: @sshumake Thanks for the review, while checking event listeners most places recommended using $scope.$on(), but since we were using $rootscope i went with that, but the former works fine.

renujhamtani:
Reviewed and merging!

PR #125, opened 2015-01-29
Changes for taking always latest version of strata from github tag

pbathia:


renujhamtani:
Reviewed !! Merging!!

@sshumake  I am merging this as we had decided to go ahead with * instead of > version

PR #126, opened 2015-01-30
Changes for PCM-397 and also handling small jshint suggestion

pbathia:


pbathia:

PR #147, opened 2015-02-27
Fix chroming errors

sshumake:
@ardasgup @renujhamtani  I am verifying and merging since this should never make it out of dev.  Please review and let me know if you see any problems.


ardasgup:
Reviewed! 
@sshumake i changed the escalation controller name so will push the same for /chromed.html

PR #154, opened 2015-03-04
don't show error on 200 response after updating case

connyay:


connyay:
Fixed by https://github.com/redhataccess/stratajs/commit/6d671a4ecce79586281a3ea44fab71cfa5794e4b

PR #163, opened 2015-03-20
changes for adding provision for making call to UDS via udsjs

pbathia:
modified:   Gruntfile.js
	modified:   app/ascension/controllers/caseQueue.js
	modified:   app/ascension/services/caseQueueService.js
	modified:   app/chromed.html
	new file:   app/common/services/udsService.js
	modified:   app/index.html
	modified:   app/loader.js
	modified:   bower.json
	modified:   karma.conf.js

### DIFF
--- a/app/common/services/udsService.js
+++ b/app/common/services/udsService.js
@@ -1,0 +1,74 @@
+'use strict';
+/*global navigator, strata, uds, angular*/
+angular.module('RedhatAccess.common').factory('udsService', [
+    '$q',
+    'RHAUtils',
+    '$angularCacheFactory',
+    function ($q, RHAUtils, $angularCacheFactory) {
+        var service = {
+            cases: {
+                list: function(uql) {
+                    var deferred = $q.defer();
+                    uds.fetchCases(
+                        function (response) {
+                            deferred.resolve(response);
+                        },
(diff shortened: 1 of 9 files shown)


pbathia:
@sshumake Please review and merge.

app/common/services/udsService.js
```diff
@@ -0,0 +1,74 @@
+'use strict';
```
sshumake: We should enable caching....however this is outside the scope of this pull request

bower.json
```diff
@@ -13,6 +13,7 @@
         "angular-ui-router": "0.2.8-bowratic-tedium",
         "angular-bootstrap": "0.10.0",
         "stratajs": "~1.1.14",
+        "udsjs":"*",
```
sshumake: I'm ok with using * at the moment, but this needs to change once udsjs will have less churn.

sshumake:
Looks good!

app/common/services/udsService.js
pbathia: yes i agree, but i was of the opinion once we start pulling actual data from UDS, we will add caching, this will allow us better testing. So i kept the parameter as it is, as we wont forget it.

PR #167, opened 2015-03-23
PCM-610 Initial MockUp

renujhamtani:
@mattnolting @sshumake @pbathia Please review and let me know the comments.

The topCases.jade ( final output is not looking good ). Please suggest some changes. I am using Leslie's mockup designs[1]

[1] https://mojo.redhat.com/docs/DOC-1019239?et=notification.mention#comment-990568

### DIFF
--- a/app/ascension/views/caseQueue.jade
+++ b/app/ascension/views/caseQueue.jade
@@ -1,5 +1,30 @@
 .page-header(rha-header='', page='')
-div
-  .row
-    section.main.col-md-8
-      h1.page-title ASCENSION DUMMY PAGE
+//-replace the ng-how wiht the security service and sfdc health check and header parameter
+div(ng-show='true')
+    .row
+        .row
+            .col-sm-12(rha-topcases='')
+        section.main.col-md-9
+            .row
+            .row
+                .col-sm-5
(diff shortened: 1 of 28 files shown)


app/ascension/views/caseQueue.jade
```diff
@@ -1,5 +1,30 @@
...
+    .row
+        .row
+            .col-sm-12(rha-topcases='')
+        section.main.col-md-9
+            .row
+            .row
+                .col-sm-5
+                    a.btn.btn-info.btn-sm(href='#/case/new'){{'Edit'|translate}}
+                    &nbsp;
+                    a.btn.btn-info.btn-sm(translate='', ng-click='', ng-disabled=''){{'Close'|translate}}
```
pbathia: what is translate inside round bracket mean ? translate should be only at one place for one dom element right ?

mattnolting: The row nesting is incorrect. Should be .row > (immediate child) .col-sm-12

mattnolting: Remove this row

renujhamtani:
I have incorporated Matt and Pranjal's comments. I am merging this for now as there are many files to be pushed and this would not hamper any other work.  So pushing this for now. Please review it as and when you get time and i will incorporate those suggestions.

PR #173, opened 2015-03-23
pcm-596

sshumake:
@renujhamtani @pbathia

### DIFF
--- a/app/cases/controllers/addCommentSection.js
+++ b/app/cases/controllers/addCommentSection.js
@@ -44,24 +44,7 @@ angular.module('RedhatAccess.cases').controller('AddCommentSection', [
                 }
                 CaseService.commentText = '';
                 CaseService.disableAddComment = true;
-                //TODO: find better way than hard code
-                var status = {};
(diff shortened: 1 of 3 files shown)


app/cases/controllers/addCommentSection.js
```diff
@@ -88,7 +71,7 @@ angular.module('RedhatAccess.cases').controller('AddCommentSection', [
                 $scope.addingComment = false;
                 $scope.progressCount = 0;
             };
-            if(!CaseService.disableAddComment){
+            if(!CaseService.disableAddComment && CaseService.commentText !== 'undefined'){
```
pbathia: I didn't understand the purpose for adding this condition, would you please explain the scenario.

sshumake: The same button submits attachments and adds comments.  I was running into a scenario where when only an attachment was being added if the comment field was blank a blank comment was being attached resulting in an error.

sshumake: @pbathia ^

pbathia: ok got it Spenser. Thanks..

renujhamtani:
Reviewed!Merging!

PR #175, opened 2015-03-25
Fix for PCM-645 unit tests Services

ardasgup:
@sshumake @renujhamtani pls review and merge


renujhamtani:
Reviewing and merging this too as I also have some of the files and would give a merge conflict.

@sshumake  Please review even if it is merged.

PR #180, opened 2015-03-26
 adding ID attribute to some div

pbathia:


pbathia:
@sshumake Please review and merge

PR #182, opened 2015-03-30
chnages for pcm-636, if solution is both pinned and handpicked, it shoul...

pbathia:
...d appear in both place.


pbathia:
@renujhamtani @sshumake Please review and merge.
Fix for PCM-636, is only change at line 51 in recommendationService.js and in jade file.

Other code I removed as, it was causing duplication of some solutions in 'Top solution' section. 
As we were pushing tmpRec to this.recommendation array. In the same array we were pushing the result from /rs/problem endpoint which was causing duplication of solutions in same array.

renujhamtani:
Reviewed! Looks good to me!

PR #189, opened 2015-03-31
Fix for PCM-422

ardasgup:
@sshumake @pbathia pls review and merge !
@sshumake once you merge the stratajs change , i will bump that, because this commit is dependent on that.


pbathia:
Reviewed..looks good to me.

PR #196, opened 2015-04-07
Pcm 688 and PCM 642

sshumake:
@pbathia @renujhamtani


sshumake:
@pbathia @renujhamtani This is my first pass at the a/b testing and the routing.  I am in the process of deciding how to best route users.

pbathia:
Looks good to me. Are we going to do this AB testing for specific period in Prod ? or just QA ?

renujhamtani:
looks good to me !! merging!!

PR #197, opened 2015-04-08
Pcm 69

renujhamtani:


ardasgup:
@renujhamtani  your changes are absolutely fine. But since i assume this is across the app for all date/time fields i would suggest to incorporate a common interface for converting all at the very source of the data. So that next time we add some date/time field somewhere, we dont have to explicitely call that converter in our controllers.
What i envisage here is, for all strata response objects containing date/time parameters, pass them through a filter(your moment lib and formatter) before returning the promise, this could be done in the strataService.js since it resides in the common module you could easily call the RHAUtils in it.
This would avoid changes at multiple levels and keep a clean jade+controller. We would also have a single point of control for all date fields in the StrataService.

sshumake:
I agree with Aritro.  Perfectly fine as is, but it would be cleaner in the strataService.

PR #201, opened 2015-04-09
pcm-665

sshumake:
@ardasgup @renujhamtani


ardasgup:
Reviewed!

renujhamtani:
@sshumake The translation needs to go in  the template.pot as this would be the only string left out. The bugzilla is already raised.  We may have to take this for the next sprint translation. Any thoughts??

PR #203, opened 2015-04-13
Pcm 69

renujhamtani:
@ardasgup @sshumake  Please review and merge. I have incorporated the changes suggested by Aritro to put the logic in strataService.js


sshumake:
Looks good @renujhamtani We will need to coordinate the adding Require JS changes to our ITOS gear when this goes live.

PR #213, opened 2015-04-16
Pcm 851

renujhamtani:
@ardasgup @pbathia Please review and merge. I have added testcases for cases/services to increase the code coverage.


renujhamtani:
@ardasgup  Wherver the caseService object or any other object is used instead of mockStrataDataService, it was because we needed not to mock strata but to use input values for caseService or any other object.

renujhamtani:
@ardasgup Incorporated ur comment! thanks!

ardasgup:
Reviewed!

pbathia:
Reviewed.

PR #215, opened 2015-04-17
Fix for PCM-689

ardasgup:
@renujhamtani @pbathia pls review and merge!


renujhamtani:
@ardasgup  We want to enable bugzillas for external customers in this issue. The code would work if the bugzilla count is not 0 so the external customers would be able to see that (if count>0) which is correct. But internal user would see the bugzillas even if the count is 0. Is this the criteria?

ardasgup:
@renujhamtani  Yes! Basically the internal users should have that BZ tab always present. If BZs are not present then message "No linked BZ". But for customers that tab and data should only be shown if there exists BZs in the payload.

renujhamtani:
Reviewed!

pbathia:
Reviewed.

PR #218, opened 2015-04-20
Fix for PCM-764

ardasgup:
@renujhamtani @pbathia pls review and merge!


pbathia:
Looks good. 
Just one thing, if we create new case, successfully created case alert will appear, and if we come back to case list page, we should remove it or not ? if yes i think we need to consider this code change.

renujhamtani:
Agreed with pranjal. I am not sure whether we have looked at the whole application for creating alerts. If yes, then we are good to go.

ardasgup:
Yes we should clear all previous messages whenever we navigate to a new page, that is my understanding.

renujhamtani:
I am merging this. Pranjal, if you have any doubt, please let aritro know.

PR #220, opened 2015-04-21
PCM-854 changes

renujhamtani:
@ardasgup @pbathia Please review


ardasgup:
Reviewed !

renujhamtani:
Merging this as i have a conflict with another issue. Pranjal, please review this even if this is merged. I can submit a new PR if there are any review comments.

pbathia:
@renujhamtani Dont you think we should use CaseService instead of this @ https://github.com/renujhamtani/redhat_access_angular_ui/blob/PCM-839/app/cases/controllers/new.js#L241

PR #231, opened 2015-04-23
changes for pcm-902

pbathia:


pbathia:

renujhamtani:
Reviewed!

pbathia:
@ardasgup Please review. I have kept the foreach loop as it is as performance is better than native for loop.
For assignment I have changed to only flag instead of full object.

pbathia:
@renujhamtani  @ardasgup Reverted the changes, as it is not working when we just try to assign some property value when we are iterating the same object.

ardasgup:
Reviewed!

PR #233, opened 2015-04-23
changes for pcm-879

pbathia:


pbathia:

renujhamtani:
Looks good to me. Testing needs to be thorough!

pbathia:
@ardasgup selectPage is not just called from Jade, it is also called from watch on reommendations, pinnedRecommendations and handpickedRecommendations. We need increment/decrement functionality from just jade based on next or previous button click. Adding conditional case is selectPage method will increase unnecessary code complexity. So i think we will just keep it simple, as already I have added scope[page] to avoid code duplication for different type of page on different section. 
Let me know your view, and then we can decide further.

renujhamtani:
Merging this as per Aritro's consent.

PR #252, opened 2015-05-06
pcm-899

sshumake:
We never filter by owner_ssoname so this is not needed.  I have commented out because I am opening a larger task in Jira to completely decouple all of our pages dependency on CaseService.  The cross-pollination of data is problematic.

@ardasgup @renujhamtani


renujhamtani:
Reviewed!

ardasgup:
Reviewed!

PR #258, opened 2015-05-07
Fix for PCM-342

ardasgup:
@sshumake @renujhamtani Pls review !
So this commit basically takes into account spenser's search parameters persistence and the 'sort by' localstorage changes.

### DIFF
--- a/app/cases/controllers/filterSelect.js
+++ b/app/cases/controllers/filterSelect.js
@@ -6,51 +6,20 @@ angular.module('RedhatAccess.cases').controller('FilterSelect', [
     'CaseService',
     'STATUS',
     'translate',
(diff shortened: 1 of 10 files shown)


sshumake:
Looks good to me.

renujhamtani:
This needs thorough testing.

renujhamtani:
@sshumake @pbathia  @ardasgup This looks like a big change and after taking this into CI, i could see small defects like "Open and Closed" support cases not working and  sort by option has "Select option" field. Can we do one thing? can we revert these changes and take this up after thorough testing next week ?? We will not push this code to prod.  After we are done pushing to prod next week, we can work on this parallel to ascension. What say??

PR #263, opened 2015-05-08
changes for pcm-923, code coverage is above 80% in all except product…

pbathia:
…Service


pbathia:
There is some redundant code, for which we can not add unit test cases and also branch coverage.
For this redundant code I am opening jira with trivial.
Average of all services is around 90%.
@renujhamtani @sshumake Please review and merge.

renujhamtani:
Looks Good!

PR #271, opened 2015-05-18
Restructuring the ascension code

ardasgup:
@renujhamtani @pbathia pls review and merge ! Mostly it is renaming, adding some more views and restructuring the parent view as per the new wireframe.


app/ascension/controllers/yourCases.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 /*jshint camelcase: false */
-angular.module('RedhatAccess.ascension').controller('TopCases', [
+angular.module('RedhatAccess.ascension').controller('YourCases', [
```
pbathia: I think topCases was not suitable than yourCases, as we are not showing all his/her cases, just top 6/7 out of that.

pbathia:
Rest all looks good.

app/ascension/controllers/yourCases.js
ardasgup: I just went ahead with the direct verbiage used on the UI, So that it is simpler to read, But if we need further arbitration we can discuss and rename.

renujhamtani:
We can rename as and when development of that particular module comes up. I am merging this.

PR #274, opened 2015-05-20
pcm-494

sshumake:
This was a major PITA that needs reviewed by everyone @ardasgup @pbathia @renujhamtani


ardasgup:
Reviewed!

renujhamtani:
Reviewed! although not understood every scenario.

pbathia:
Reviewed. We definitely need through testing for internal user validations.

PR #276, opened 2015-05-21
Fix for PCM-963

ardasgup:
@sshumake @renujhamtani pls review and merge!


renujhamtani:
@sshumake @ardasgup :
The Travis CI build failed because of a test case which Spenser is going to fix. Other than that, Aritro's code looks good. 

I am merging this. Spenser, please review this even if this is merged as i need these changes.

PR #278, opened 2015-05-25
Fix for PCM-865

ardasgup:
@renujhamtani @pbathia please review and merge ! This is an adhoc change as i am not very convinced how the flags would work, have some better ideas, but for now will keep it simple.


renujhamtani:
The travis CI build is failing because of one unit test case(which spenser was going to comment out. ). Other than that, the code looks good.

pbathia:
Reviewed.

PR #279, opened 2015-05-26
Initial commit which also has the change for test case commenting

renujhamtani:
@ardasgup @pbathia Please review. I have much more work to do in this but i am creating a PR to avoid merge conflicts as these are very common files.

### DIFF
--- a/app/ascension/controllers/caseDiscussion.js
+++ b/app/ascension/controllers/caseDiscussion.js
@@ -3,9 +3,91 @@
 /*jshint camelcase: false */
 angular.module('RedhatAccess.ascension').controller('CaseDiscussion', [
     '$scope',
-    function ($scope) {
+    '$timeout',
+    'CaseDiscussionService',
+    'udsService',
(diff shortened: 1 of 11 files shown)


ardasgup:
Reviewed!

app/ascension/controllers/caseDiscussion.js
```diff
@@ -3,9 +3,91 @@
...
+    'securityService', //replace by UDSService
+    '$stateParams',
+    'AlertService',
+    '$modal',
+    '$location',
+    '$anchorScroll',
+    'RHAUtils',
+    'EDIT_CASE_CONFIG',
+    'AUTH_EVENTS',
+    'CASE_EVENTS',
```
pbathia: I will suggest remove extra parameters which are not used. As those remains as it is and creates code quality standard issue when we are running sonar analysis.

renujhamtani:
Would incorporate these comments later as this is WIP,

PR #281, opened 2015-05-27
Revert "Changes for PCM-342"

sshumake:
Reverts redhataccess/redhat_access_angular_ui#272


sshumake:
@ardasgup @pbathia @renujhamtani Revert of pcm-342

renujhamtani:
@sshumake Yup the test failures are the ones which were specifically changed for PCM-342. I think we should not fix this? as again next week we would merge this and again would have to revert the test failures.

sshumake:
I agree we should not fix this.

PR #282, opened 2015-05-27
Revert "Revert "Changes for PCM-342""

sshumake:
Reverts redhataccess/redhat_access_angular_ui#281


renujhamtani:
@sshumake @pbathia @ardasgup I am not merging this until we get a confirmation from QA that everything is working good on QA with tag 1.0.3 . If something is not working and which needs to be fixed, we can take that up first.

sshumake:
@pbathia @ardasgup @ardasgup  Remerging so unit tests start passing again.

PR #291, opened 2015-06-02
changes for user role logic, as one ssoname can have multiple users a…

pbathia:
…nd hence many roles


pbathia:

ardasgup:
Reviewed!

PR #293, opened 2015-06-02
removing redundant code

pbathia:


pbathia:
@ardasgup @sshumake please review and merge. If you guys suggest I can remove that code instead of commenting out.

sshumake:
This code will be refactored when we move fully to solr.  I'l OK with it just being commented out.

ardasgup:
Reviewed!

PR #305, opened 2015-06-08
Pcm 973

sshumake:
@renujhamtani @pbathia


app/chromed.html
```diff
@@ -1,13 +1,14 @@
...
-    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <!--[if lte IE 8]>
+    <meta http-equiv="x-ua-compatible" content="IE=Edge"/>
+    <![endif]-->
+    <meta http-equiv="Content-Type" content="text/html; charset=US-ASCII" />
     <link href="/chrome_themes/nimbus/css/chrometwo.css" rel="stylesheet" />
     <script type="text/javascript" src="/webassets/avalon/j/lib/require.js"></script>
     <title></title>
     <meta name="description" content="">
-    <meta name="viewport" content="width=device-width">
```
pbathia: I think this line is for responsiveness of site, do we need to remove this for cache stuff?

renujhamtani:
Looks good to me!

app/chromed.html
sshumake: This is used for mobile development.  I have refactored this file to match our prod file which does not have this line

pbathia: Ok.

PR #308, opened 2015-06-08
Monkey patch

sshumake:
@renujhamtani @pbathia 

This is a definite monkey patch but the browser when copying and pasting has a bug where it allows more characters in the text box even though you set a hard limit.  This is as good as it gets for now.


pbathia:
Reviewed.

renujhamtani:
Looks good!

PR #309, opened 2015-06-08
pcm-1036

sshumake:
@ardasgup @renujhamtani


renujhamtani:
@sshumake  This is quite a easy fix .I think many customers complained of this. Even saw ur comment on facebook for redhat support. If this is the fix, this is awesome. Thanks Spenser!!

ardasgup:
Reviewed!

PR #310, opened 2015-06-08
Gs4 beta

sshumake:
@ardasgup @pbathia @renujhamtani Please look at and make sure there are sane changes


pbathia:
Reviewed.

renujhamtani:
@sshumake Looks a big change. How do we test this? I could not get most of the code:)

ardasgup:
The changes are sane :)  who will be testing it ? Don't think we @pune have access to the GS4 instance.

PR #313, opened 2015-06-10
PCM-1046 changes

renujhamtani:
@mattnolting @sshumake Please review the changes for PCM-1046. Was there a better way to do this?
The label "Send Email Notifications to" is not showing up in emailNotifySelect.jade because it cannot find .label. So i have added .label copied from _config-options.scss to _case-discussion.scss. Is this the right way to do this? Can we reuse label instead of duplicating code? I am not sure on how to do that..

### DIFF
--- a/app/assets/sass/components/_case-discussion.scss
+++ b/app/assets/sass/components/_case-discussion.scss
@@ -21,4 +21,13 @@
 	.tab-pane {
 		padding: 0;
 	}
+
+    .label {
+        display: block;
+        padding: 0;
+        font-weight: 700;
+        text-align: left;
+        @include font-size(13px);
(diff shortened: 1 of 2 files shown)


renujhamtani:
@ardasgup @pbathia Please review this. I have asked Matt also to review this. This might be a temporary solution until matt confirms. But we need to fix this on UI as this is a major regression bug as the label is not showing up.

ardasgup:
Reviewed!

pbathia:
Looks good to me.

mattnolting:
@renujhamtani the css should, and now does, support the semantic standard label. i've updated the css and jade.

PR #320, opened 2015-06-16
Fix for PCM-1032

ardasgup:
@sshumake  @renujhamtani  please review and merge !

### DIFF
--- a/app/ascension/controllers/caseDetails.js
+++ b/app/ascension/controllers/caseDetails.js
@@ -7,10 +7,39 @@ angular.module('RedhatAccess.ascension').controller('CaseDetails', [
     'AUTH_EVENTS',
     function ($scope, CaseDetailsService, securityService, AUTH_EVENTS) {
     	$scope.CaseDetailsService = CaseDetailsService;
+        $scope.showEditCase = false;
 
     	$scope.init = function () {
             //CaseDetailsService.getCaseDetails(CaseDetailsService.cases[0]);
         };
+
+        $scope.edit = function() {
+            $scope.showEditCase = true;
+            CaseDetailsService.fetchProducts();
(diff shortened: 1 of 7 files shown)


app/ascension/controllers/caseDetails.js
```diff
@@ -7,10 +7,39 @@ angular.module('RedhatAccess.ascension').controller('CaseDetails', [
...
     function ($scope, CaseDetailsService, securityService, AUTH_EVENTS) {
     	$scope.CaseDetailsService = CaseDetailsService;
+        $scope.showEditCase = false;
 
     	$scope.init = function () {
             //CaseDetailsService.getCaseDetails(CaseDetailsService.cases[0]);
         };
+
+        $scope.edit = function() {
+            $scope.showEditCase = true;
```
sshumake: Products, severities, and statuses should be fetched on page load.

renujhamtani:
Reviewed!!

ardasgup:
@renujhamtani pls review and merge, incorporated Spenser's review comment.

app/ascension/controllers/caseDetails.js
renujhamtani: @sshumake I do not agree to this as ,if in ascension,if the associate does not click on edit, there is  no need to fetch products, severities and statuses. These all should be loaded only after the user clicks on "edit". Anyways, I was late in commenting. Aritro has already made the changes. So merging this.

sshumake: @renujhamtani I'm happy to discuss more.  It should be loaded initially because it's an asynchronous call that happens in the background.  If we don't load initially we will make the user unnecessarily wait for the call to happen when the click on edit.

PR #328, opened 2015-06-22
Remove A/b routing logic

renujhamtani:
@ardasgup @pbathia @sshumake Please review and merge.
(I have informed Swapna to change her automation suite based on these changes once this is merged on CI).


renujhamtani:
@ardasgup The new Router was just a a/b routing for new page. In future, we can have it at any place like list, edit etc. I thought it would have been better to remove all those not needed currently. I am open to options.
@sshumake Can you comment?

ardasgup:
@renujhamtani agreed this was for new case page, but this file specifically was to set up the A/B test omniture condition and the routing logic. Maybe Spenser can take a call.

sshumake:
@ardasgup  @renujhamtani I'm fine with fully removing for now.  However, as apart of implementing the next test we need to discuss how to make that file more scalable so we are not adding/removing with every change.

pbathia:
Reviewed.

renujhamtani:
I am merging this now.

PR #330, opened 2015-06-22
Fix for PCM-770

ardasgup:
@sshumake @renujhamtani pls review and merge!

### DIFF
--- a/app/cases/services/searchCaseService.js
+++ b/app/cases/services/searchCaseService.js
@@ -188,7 +188,8 @@ angular.module('RedhatAccess.cases').service('SearchCaseService', [
                 if(sortField === "owner"){
                     sortField = "contactName";
                 }
-                cases = strataService.cases.search(this.caseParameters.status, caseOwner, this.caseParameters.group, this.caseParameters.searchTerm, sortField, CaseService.filterSelect.sortOrder, this.start, this.count, null, null).then(angular.bind(that, function (response) {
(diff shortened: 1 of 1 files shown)


renujhamtani:
looks good to me!

renujhamtani:
@sshumake Please review this change. We need to merge this only after stratajs is merged and version is bumped and updated.

app/cases/services/searchCaseService.js
```diff
@@ -188,7 +188,8 @@ angular.module('RedhatAccess.cases').service('SearchCaseService', [
                 if(sortField === "owner"){
                     sortField = "contactName";
                 }
```
sshumake: @ardasgup I do not think we want to pass the account number. It is my understanding Solr filters based on account number by default EXCEPT for internal users.  This change will break PCM for internal users.  Why has this been added?  I do not see it as apart of the Jira.  By adding things not explicitly in Jiras we risk introducing regressions that could not be caught by test.

ardasgup: @sshumake i was not aware of the default nature of the account number for internal users. Just tested and it seems to be the case, I had added that because Charles had it in his query when showing how to do the search and i thought ours was having issues because of not passing the same. Sorry for blindly following his query :/
I will remove it.

ardasgup:
@renujhamtani have made the changes as per @sshumake review pls check and merge !

renujhamtani:
Looks good! Merging!

PR #334, opened 2015-06-24
NEP display fields

renujhamtani:
@ardasgup @sshumake  please review and merge. This is for display of NEP fields in ascension.

### DIFF
--- a/app/ascension/services/caseDetailsService.js
+++ b/app/ascension/services/caseDetailsService.js
@@ -59,7 +59,30 @@ angular.module('RedhatAccess.ascension').service('CaseDetailsService', [
                         }
                     }
                 }
-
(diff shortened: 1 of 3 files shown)


renujhamtani:
@ardasgup Thanks for catching the translation one. The translation PR is pending and caseDetailsService has the latest changes which i had done for the translation issue. I will get a merge conflict for this . Hence i will make sure that "Yes" and "No" are also translated. Because of translation, it would be better if i keep this in the .js file and not the jade file using the ternary operator. I will merge this if you are fine with this change.

@sshumake  I am merging this(your review id not yet done). I have the QA tag to be pushed. So want this to go in that tag. Please review this even if this is merged.

ardasgup:
@renujhamtani i am ok with the translations being handled in .js file.

PR #335, opened 2015-06-24
Fix for PCM-572

ardasgup:
@renujhamtani @pbathia please review and merge!

### DIFF
--- a/app/ascension/controllers/caseHistory.js
+++ b/app/ascension/controllers/caseHistory.js
@@ -2,7 +2,19 @@
 /*jshint camelcase: false */
 angular.module('RedhatAccess.ascension').controller('CaseHistory', [
     '$scope',
-    function ($scope) {
+    'CaseDetailsService',
+    'RHAUtils',
+    '$sce',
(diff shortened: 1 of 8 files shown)


app/ascension/controllers/caseHistory.js
```diff
@@ -2,7 +2,19 @@
...
+    'RHAUtils',
+    '$sce',
+    function ($scope, CaseDetailsService, RHAUtils, $sce) {
+    	$scope.CaseDetailsService = CaseDetailsService;
 
+    	$scope.parseHistoryHtml = function (text) {
+            var parsedHtml = '';
+            if (RHAUtils.isNotEmpty(text)) {
+                var rawHtml = text.toString();
+                parsedHtml = $sce.trustAsHtml(rawHtml);
```
pbathia: does some html stuff come as a part of output text ?

pbathia:
Rest all code looks good.

app/ascension/controllers/caseHistory.js
ardasgup: Yeah the entire outputText is a html literal, similar to the attachment artifact so need to parse it.

PR #337, opened 2015-06-25
Fix for PCM-950 and PCM-953

avijra:
@renujhamtani @sshumake please review and merge


avijra:
on discussion with @ardasgup found issues with PCM-950 hence closing the pull request

PR #342, opened 2015-06-29
changes for pcm-578, this is first pass until we get mockup from Leslie

pbathia:


pbathia:
On strata/ccm side there is some wrong on mapping done for case subject and case summary. So there is mismatch between how data is returned from UDS and strata for subject and case summary.

ardasgup:
Reviewed!

PR #344, opened 2015-06-30
pcm-1087

sshumake:
@ardasgup @renujhamtani


renujhamtani:
@sshumake WE do not commit the html file(app/search/views/searchStackedSolutions.html). This html file was already in git. We need to remove this html from git as it causes confusion.

ardasgup:
Reviewed! except that gitignore part for .html files under /search/views/

renujhamtani:
I am merging this. We can remove the html file in next iteraiton.

PR #353, opened 2015-07-01
pcm-1094

sshumake:
@ardasgup @pbathia @renujhamtani


ardasgup:
@sshumake not sure if group.number = -1 can have group.is_default = true :/ but if you have tested maybe its true.
Reviewed!

renujhamtani:
Reviewed!

PR #357, opened 2015-07-02
Fix or PCM-953

avijra:
@renujhamtani @ardasgup  Please review


ardasgup:
@renujhamtani this looks like an old file, does not have your translation changes, however marking it reviewed as the fix is ok.
Reviewed!

renujhamtani:
@sshumake I don't want to merge this unless we go live on monday. Because "Close case" is on the main  page and that would show it as untranslated and i don't want that as I tested this completely for all strings translated.:)

Should we merge this once we go live?

sshumake:
@renujhamtani  yes since its a very minor change let's wait for translations until we merge

PR #365, opened 2015-07-08
Pcm 1072

renujhamtani:
@ardasgup @pbathia Please review. This only does not have customer name info which UDS needs to provide. Other than that, everything is done.


app/ascension/controllers/caseDiscussion.js
```diff
@@ -132,14 +132,16 @@ angular.module('RedhatAccess.ascension').controller('CaseDiscussion', [
...
-                if (RHAUtils.isNotEmpty(comment.resource.body)) {
-                    var rawHtml = comment.resource.body.toString();
-                    parsedHtml = $sce.trustAsHtml(rawHtml);
-                }
-            } else if (comment.resource.text !== undefined) {
-                if (RHAUtils.isNotEmpty(comment.resource.text)) {
-                    parsedHtml = $filter('linky')(comment.resource.text,'_blank');
+            if(comment.resource) {
+                if (comment.resource.body !== undefined) {
+                    if (RHAUtils.isNotEmpty(comment.resource.body)) {
```
pbathia: This is double checking with same condition, as isNotEmpty checks for undefined value also.

app/ascension/controllers/caseDiscussion.js
```diff
@@ -132,14 +132,16 @@ angular.module('RedhatAccess.ascension').controller('CaseDiscussion', [
...
+                        var rawHtml = comment.resource.body.toString();
+                        parsedHtml = $sce.trustAsHtml(rawHtml);
+                    }
+                } else if (comment.resource.text !== undefined) {
+                    if (RHAUtils.isNotEmpty(comment.resource.text)) {
```
pbathia: Same as above. We can avoid 1 if condition

app/ascension/controllers/caseSearch.js
```diff
@@ -2,11 +2,14 @@
...
     '$scope',
+    '$rootScope',
     'CaseDetailsService',
-    function ($scope, CaseDetailsService) {
+    'TOPCASES_EVENTS',
+    function ($scope,$rootScope, CaseDetailsService,TOPCASES_EVENTS) {
     	$scope.caseNumber = '';
 
     	$scope.searchCases = function(){
+            $rootScope.$broadcast(TOPCASES_EVENTS.topCaseFetched);
```
pbathia: why you have added this ?

renujhamtani:
@pbathia Incorporated the review comments. Please review

pbathia:
Reviewed.

app/ascension/controllers/caseSearch.js
renujhamtani: Whenever we click on top cases (any case)..the case details get loaded initially and then discussion section. When we search a case, the case details does not get loaded if we are in the discussion section. Hence discussion section shows the stale information. I am now defaulting this to case details section everytime so that the data is refreshed. This would change in future as we might retain the same section. But i did not want any stale data to come up in short term live. This applies to all sections. (Case history etc)

ardasgup:
Reviewed!

PR #366, opened 2015-07-08
Pcm 992

renujhamtani:
@ardasgup @pbathia Please review. This is a defect in case discussion when u toggle between different elements like discussion, history, esclations etc and then come back to discusisons, the screen was not updating for comments/attachements.


pbathia:
Reviewed.

ardasgup:
Reviewed!

PR #374, opened 2015-07-10
Fixes on case discussion and account note deletion

renujhamtani:
@pbathia @ardasgup Please review. These contain the changes for "Account notes deletion" as its not in this sprint and was half baked and giving errors. The other changes are to the case discussion section to show the correct dates and names after Spenser changes. Still the sorting is not working. Hope Spenser would figure this out or I would if i get time.


ardasgup:
Reviewed!

pbathia:
Reviewed.

PR #390, opened 2015-07-15
compress image

sshumake:
IE8-- @renujhamtani @pbathia @ardasgup


pbathia:
Reviewed.

renujhamtani:
why the tag IE 8 is written in comments? we do not support ie for ascension..??

PR #402, opened 2015-07-20
Pcm 1116

sshumake:
@renujhamtani @pbathia @ardasgup @avijra first pass?

PR #404, opened 2015-07-21
changes for pcm-1127

pbathia:


pbathia:
@ardasgup @renujhamtani Please review and merge.

I don't have any changes in caseService.js, but due to yesterday's commit few and then merge conflicts resolution got some some indentation changes in that file.

renujhamtani:
Looks good. Need to manually merge this once ascension PR 1112(https://github.com/redhataccess/redhat_access_angular_ui/pull/402) is merged. The file which needs to manually merged is  app/ascension/services/caseDetailsService.js

app/ascension/services/caseDetailsService.js
```diff
@@ -274,28 +274,47 @@ angular.module('RedhatAccess.ascension').service('CaseDetailsService', [
                             topCases.sort(function (a, b) {
                                 return b.resource.collaborationScore - a.resource.collaborationScore;
                             });
-                            if(!RHAUtils.isObjectEmpty(self.enggBackLogCases)) {
-                                topCases = topCases.slice(0,self.yourCasesLimit - 1); //if engineering backlog case is found, we will append at end and slice top 5 cases
-                                topCases.push(self.enggBackLogCases[0]);
-                            }
+                            if(this.foundEnggBacklogRole) {
```
ardasgup: @pbathia this looks confusing to me , i see there are two broadcasts for caseLoad event. Not sure why? Also the locked case filtering should happen after the engineeringBacklog filter completes, not sure from code if that is taken care. I checked the raw file and it somehow doesn't look proper to me. Have you tested all the scenarios ?

pbathia: @ardasgup Yes, as some users wont have engineering backlog role, in that case also broadcast event should happen.
And also if you check the JIRA i already mentioned all possible scenarios I can think of.

ardasgup: @pbathia i meant one broadcast from engineeringbacklog event and other from lockedcase event. can these not be combined or something, wont there be two fetch case details then. Am just saying looking at code.

pbathia: @ardasgup  for combining, we wont be able to combine in all cases, as engineering backlog role is optional. So it should work in all cases. 
As discussed over IRC, we need to combine all promises in array and use $.all(promises) before firing broadcast event. We can take this in hardening task, as this needs proper testing.

ardasgup:
Reviewed!

renujhamtani:
@pbathia please create a hardening task for this. I am merging this for now.

PR #410, opened 2015-07-24
Pcm 1092

sshumake:
@renujhamtani @ardasgup @avijra


ardasgup:
@sshumake these look good, i see lot of elaborate changes, we will need to get it tested whole heartedly specially the recommendations and localstorage stuff.
Also, i may be trivializing this, but i was thinking we should have a better framework for doing these. Avoid changing existing files as much as possible. maybe just make minor adjustments. And have the A/B route or component or page configurable to some point. I am sure you must have thought that out , so just wondering if that is not possible at all.

renujhamtani:
Looks Good!

PR #443, opened 2015-09-02
Code changes for pcm 1270

gandhikeyur:
@renujhamtani @ardasgup Please review


renujhamtani:
Looks good!

ardasgup:
@gandhikeyur i didn't exactly get the change, it looks you just moved out the span component from its parent one ?

gandhikeyur:
@ardasgup Yes, that was the only change that was needed for the fix. The close button was implemented as a part of the href url, which was creating the issue as clicking on the close button tried to open the attachment in the url.

ardasgup:
Reviewed!

PR #453, opened 2015-10-08
PCM-952

jtrantin:
Changed rem units for the button's font size and padding to pixels, because we need to match the input size, which is specified in pixels. Made more sense than approximating the pixels needed via rem units.
@renujhamtani @vrathee Please review.


renujhamtani:
Looks good!

vrathee:
Looks Good.

PR #455, opened 2015-10-09
PCM-945

jtrantin:
@renujhamtani @vrathee Please review.


renujhamtani:
@jtrantin I think this does not look correct. Now only the internal users who have bugzillas would be able to see it. What about external users? I will ask @ardasgup for his views. May be I am wrong but just want to be sure this is correct

jtrantin:
@renujhamtani It's possible I have misinterpreted the original condition, but looking at it it is not very clear. Hopefully Aritro can shed more light on this.

ardasgup:
@jtrantin  yeah Renu is correct this change will bar the customers (external) from seeing the BZ tab at all.
But lets hold on a bit as i think PCM-945 should not be a bug at all, we had already discussed this while implementing. Lets wait on our QA's ack. Have already informed him.

renujhamtani:
@jtrantin I am closing without merging this as this was not a bug as agreed by the tester.

PR #461, opened 2015-10-19
PCM-1406

jtrantin:
@renujhamtani @ardasgup Please review.

### DIFF
--- a/app/cases/controllers/severitySelect.js
+++ b/app/cases/controllers/severitySelect.js
@@ -9,7 +9,12 @@ angular.module('RedhatAccess.cases').controller('SeveritySelect', [
     	});
 
         $scope.toggleDetails = function (severity, event) {
-            event.stopPropagation(); // we don't want to toggle severity
+            if(event.stopPropagation) { // we don't want to toggle severity
+                event.stopPropagation();
+            } else { // for IE8+
+                event.returnValue = false;
+                event.cancelBubble = true;
+            }
         	$scope.openedDetails[severity.name] = !$scope.openedDetails[severity.name];
         };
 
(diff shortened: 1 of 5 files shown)


ardasgup:
Reviewed! However i don't see any check for IE browser, will this be de facto for all views/browsers ?

jtrantin:
I believe me and Renu agreed that since the visual appearance is so close to the existing implementation, we might as well not duplicate code and implement this for all browsers.

app/cases/controllers/severitySelect.js
```diff
@@ -3,15 +3,20 @@ angular.module('RedhatAccess.cases').controller('SeveritySelect', [
...
     	// INIT
-    	$scope.openedDetails = {};
+        $scope.openedDetails = {};
     	angular.forEach($scope.severities, function(severity) {
     		$scope.openedDetails[severity.name] = false;
     	});
 
-        $scope.toggleDetails = function (severity) {
+        $scope.toggleDetails = function (severity, event) {
+            event.stopPropagation(); // we don't want to toggle severity
```
ardasgup: Also just verify if this particular jquery works for all browsers, in past we have faced issues with some event handlers not working in older versions of IE. :/

ardasgup:
@jtrantin ack

app/cases/controllers/severitySelect.js
jtrantin: Correct, this will not work in IE8.

jtrantin:
@renujhamtani @ardasgup Pushed the fix as described in PCM-1406.

renujhamtani:
Reviewed! Looks good! Waiting for @ardasgup to review

renujhamtani:
@ardasgup Please review even if its merged as deepak and Swapna need to test this on CI

ardasgup:
Reviewed!

PR #464, opened 2015-10-23
PCM-927

jtrantin:
@gandhikeyur @vrathee Please review.

### DIFF
--- a/app/cases/services/attachmentsService.js
+++ b/app/cases/services/attachmentsService.js
@@ -129,7 +129,7 @@ angular.module('RedhatAccess.cases').service('AttachmentsService', [
                                 attachment.last_modified_time = RHAUtils.formatDate(lastModifiedDate, 'hh:mm A Z');
                                 attachment.published_date = RHAUtils.formatDate(lastModifiedDate, 'MMM DD YYYY');
                                 attachment.published_time = RHAUtils.formatDate(lastModifiedDate, 'hh:mm A Z');
-                                AlertService.addSuccessMessage(gettextCatalog.getString('Successfully uploaded attachment {{attachmentFileName}} to case {{caseNumber}}',{attachmentFileName:attachment.file_name,caseNumber:caseId}));
+                                AlertService.addSuccessMessage(gettextCatalog.getString('Successfully uploaded attachment {{attachmentFileName}} to case {{caseNumber}}',{attachmentFileName:attachment.file_name.substring(0,80),caseNumber:caseId}));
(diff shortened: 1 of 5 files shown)


jtrantin:
Ignore this, forgot to check the tests

jtrantin:
Tests fixed, you can review now.

app/cases/services/attachmentsService.js
```diff
@@ -129,7 +129,7 @@ angular.module('RedhatAccess.cases').service('AttachmentsService', [
                                 attachment.last_modified_time = RHAUtils.formatDate(lastModifiedDate, 'hh:mm A Z');
                                 attachment.published_date = RHAUtils.formatDate(lastModifiedDate, 'MMM DD YYYY');
                                 attachment.published_time = RHAUtils.formatDate(lastModifiedDate, 'hh:mm A Z');
-                                AlertService.addSuccessMessage(gettextCatalog.getString('Successfully uploaded attachment {{attachmentFileName}} to case {{caseNumber}}',{attachmentFileName:attachment.file_name,caseNumber:caseId}));
+                                AlertService.addSuccessMessage(gettextCatalog.getString('Successfully uploaded attachment {{attachmentFileName}} to case {{caseNumber}}',{attachmentFileName:attachment.file_name.substring(0,80),caseNumber:caseId}));
```
vrathee: Since strata will eventually truncate the long file name we should do it at our end and then send the file name..
In current scenario I think we will lose the file extension which is important to us.
Do share your views on this.

jtrantin: Keeping the extension might be a good idea, I will investigate this a bit more. Anyway, there is a relevant IE issue, when a whole path is used instead of just a filename (see case 01547298), thus we would need to have separate code for IE as we should aim for equivalent behaviour across all browsers. 
Due to that and the fact the suggestion doesn't exactly correspond with the issue's AC, I am considering whether such change is not out of scope of PCM-927 as it was originally intended. @sshumake could you provide your view, whether we should implement the suggestion from Vikas or if it should be a part of another issue?

jtrantin:
As discussed with Vikas, we are keeping only the CSS changes. @vrathee @gandhikeyur Please review.

renujhamtani:
Looks good! I will merge this only once we are done with prod push

jtrantin:
After all we won't need to change the common module, so I added a commit covering the changes to this PR. Please review the last commit.

gandhikeyur:
Reviewed.

PR #482, opened 2015-12-03
PCM 1583 - calling product.getversions only if product is not undefined

gandhikeyur:
@sshumake @renujhamtani Please review


renujhamtani:
Looks good!

sshumake:
@gandhikeyur looks good, how does the UI behave?  Does the product and version default to nothing?

renujhamtani:
@sshumake yup . the UI was behaving correct on "view more". So its behaving the same way. Product and version default to nothing

PR #485, opened 2015-12-08
Pcm 1557

sshumake:
@renujhamtani @gandhikeyur @vrathee


vrathee:
@sshumake We have a "typeahead" component in angular-bootstrap ( https://angular-ui.github.io/bootstrap/ ) which we can use here instead of handling all the typeahead logic ourselves.
We already use this component to search users in ascension.

vrathee:
Looks like its not a typeahead, Reviewed !

gandhikeyur:
Reviewed

PR #486, opened 2015-12-10
Leslies changes

sshumake:
https://projects.engineering.redhat.com/browse/PCM-1477
https://projects.engineering.redhat.com/browse/PCM-1480

@jtrantin - Thoughts on changes for radio buttons?  I couldnt get it working without requiring case service
@vrathee

### DIFF
--- a/app/assets/sass/components/_case-new.scss
+++ b/app/assets/sass/components/_case-new.scss
@@ -13,7 +13,7 @@
 			margin: 5px 0 0;
 		}
 	}
-	#rha-case-more-options{
+	.rha-case-more-options{
 		label{
 			margin: 15px 0 0;
 		}
--- a/app/assets/sass/components/_case-new.scss
+++ b/app/assets/sass/components/_case-new.scss
@@ -1,0 +1,30 @@
+.case-new {
+	.config-options{
+		label{
+			margin: 15px 0 0;
+		}
+		input{
+			margin: 5px 0 0;
+		}
+		div{
+			margin: 5px 0 0;
+		}
+		textarea{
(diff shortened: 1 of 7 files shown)


jtrantin:
@sshumake This is probably the problem I encountered when first implementing this. My ultimate goal was to keep the component as reusable as possible. Unfortunately, using CaseService inside the component makes it not reusable at all.
I did a quick test and confirmed it is not possible to use severityModel from the component's scope bound to CaseService. However, I am not sure why or how to fix it and keep the component reusable.

Just a quick tip, if you add a label for the radio input and put the severity icons inside, it will be possible to toggle the severity by clicking on those icons, which are much bigger and have a responsive feel already. Just beware of possible styling issues.

sshumake:
@jtrantin I took a second shot at fighting with the scope variables in the directive.  If I bind the the kase everything works .... Not sure why we can't bind to severity.  Still a little funky but at least its still reusable.

You usually see this problem when you are binding to a variable and not an object....but severity is def an object, so I have no idea whats going on.

I fought with CSS and was defeated by the input inside the label so unless you want to try this is how it'll stay for now :)

vrathee:
@sshumake Don't know the exact context here.. But, If we need 2 way binding for 'severityChange' in directive we can use '=' instead of '&' , i guess that should help.

sshumake:
@vrathee no that's not the variable we are having troubles with, binding CaseService.kase.severity which we were binding with = is the problem

jtrantin:
@sshumake No, I tried to fight the CSS too some time ago, so let's leave it like that.

PR #507, opened 2016-01-21
Pcm header service

vrathee:
@renujhamtani @ardasgup Please Review.

The "var breadcrumbs" changes were made because there is some existing problem in loader.js and other places due to which I was unable to run PCM on local.


vrathee:
@renujhamtani The common's version in bower is 15. So, we need to first tag commons and then PCM.

ardasgup:
Reviewed!

PR #512, opened 2016-02-16
pcm-1804

sshumake:
@renujhamtani @gandhikeyur @vrathee @jtrantin

### DIFF
--- a/app/loader.js
+++ b/app/loader.js
@@ -55,7 +55,7 @@
     }
     // Once all modules have loaded bootstrap it
     jq.when.apply(jq, dfds).always(function() {
-      breadcrumbs = [
+      var breadcrumbs = [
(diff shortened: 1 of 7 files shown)


app/loader.js
```diff
@@ -55,7 +55,7 @@
     }
     // Once all modules have loaded bootstrap it
     jq.when.apply(jq, dfds).always(function() {
-      breadcrumbs = [
+      var breadcrumbs = [
```
renujhamtani: var breadcrumbs?? is this working?? This works on local but does not work on CI and QA as per me..the breadcrums fail..

sshumake: @renujhamtani  As we discussed in scrum loader.js is not built into our output.  I am merging this PR.  If anyone has concerns please add a comment.

PR #513, opened 2016-02-18
PCM-1693 code changes

gandhikeyur:
@renujhamtani Please review.
This should work for both cases : With and with out local storage cache.


renujhamtani:
Looks good!

PR #514, opened 2016-02-19
PCM-1810 code changes

gandhikeyur:
@sshumake There were a couple of things broken in the equalWidths file and that was breaking the redirect to edit case page after the case is newly created. I have done the changes, please verify as you were working on it recently.


sshumake:
Looks good.

PR #516, opened 2016-02-26
PCM-1720 fix for running PCM locally

vrathee:
@gandhikeyur @renujhamtani Please Review.
The chromed.html is used locally and thus this dependency will not affect any other environment.


gandhikeyur:
Looks good to me.

PR #525, opened 2016-03-14
PCM-1926 code changes

gandhikeyur:
The root cause of this issue lied in the assignment of CaseService.users to $scope.usersOnAccount and angular.copy solved the matter. The other file change is just for sorting it according to full name.
@sshumake @jtrantin Please review


sshumake:
Looks good to me

jtrantin:
Reviewed

PR #526, opened 2016-03-17
PCM-776: Added case number highlight

jtrantin:
@gandhikeyur Please review.
I talked with spenser and we decided that adding some hash symbol to indicate case number reference is unfeasible, due to the number of existing cases and the fact that customers would need to be forced to use it. So we decided to proceed with the regex solution; the priority is to prefer non-highlighted occurrences rather than having false positives. Let me know what you think.


gandhikeyur:
@vrathee Could you please check this as well

vrathee:
Reviewed!

gandhikeyur:
Looks good to me. Checked for a few scenarios in online regex tester and it worked as expected.
Sorry for the late review.

PR #529, opened 2016-04-20
PCM-2051: Added angular JS exception handler

jtrantin:

### DIFF
--- a/app/cases/cases.module.js
+++ b/app/cases/cases.module.js
@@ -56,7 +56,8 @@ angular.module('RedhatAccess.cases', [
     'isPCM': false
 }).config([
     '$stateProvider',
-    function($stateProvider) {
+    '$provide',
+    function($stateProvider, $provide) {
         $stateProvider.state('edit', {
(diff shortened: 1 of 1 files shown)


renujhamtani:
@jtrantin  Is this working on cukes??

renujhamtani:
@jtrantin  as far as i remember, we had removed similar functionality because of some defects*i am not remembering exactly what)..

vrathee:
Reviewed!

jtrantin:
@renujhamtani Tested with Cukes locally and it is working fine. The code that was present in the past was doing something a bit different, it was only re-throwing the errors from angular exception handler to the global exception handler. However, now we are not able to access the globally uncaught JS errors altogether (Watir and cucumber do not allow that now with newest FF), so we need to store the errors locally in the object window.errors, so the automation tests can check and read the errors anytime.

renujhamtani:
Looks good!

gandhikeyur:
Reviewed

PR #530, opened 2016-04-27
PCM-1979: Added account bookmark page + Fix for PCM-2700

jtrantin:
@renujhamtani @vrathee @gandhikeyur Please review.

### DIFF
--- a/app/cases/services/accountBookmarkService.js
+++ b/app/cases/services/accountBookmarkService.js
@@ -1,0 +1,81 @@
+'use strict';
+angular.module('RedhatAccess.cases').service('AccountBookmarkService', [
+    'securityService',
+    '$rootScope',
+    'AUTH_EVENTS',
+    'AccountService',
+    'RHAUtils',
+    'ACCOUNT_EVENTS',
+    'strataService',
+    '$timeout',
+    function (securityService, $rootScope, AUTH_EVENTS, AccountService, RHAUtils, ACCOUNT_EVENTS, strataService, $timeout) {
+        this.loading = true;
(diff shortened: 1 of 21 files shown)


app/cases/services/accountBookmarkService.js
```diff
@@ -0,0 +1,81 @@
...
+        this.removeBookmark = function (account) {
+            var user = securityService.loginStatus.authedUser;
+            return strataService.accounts.removeBookmark(account.number, user.sso_username).then(angular.bind(this, function () {
+                this.bookmarkedAccounts = this.bookmarkedAccounts.filter(function (bAcc) {
+                    return bAcc.number != account.number;
+                });
+            }));
+        };
+
+        if(securityService.loginStatus.isLoggedIn) {
```
vrathee: I am a bit skeptical about initializing service from within the service. That's why you needed the manual digest. I guess it would be better if we invoke it from one of the controller.

jtrantin: The thing is this service is used from multiple different pages, so we would need to add the initialization into multiple controllers across PCM in order to get initialized properly.
That is why I thought this is a good idea, the service initializes itself when possible and the controllers just injects it and use it and do not have to make sure it is set up properly. Similarly as the securityService, where you also do not have to make sure it is properly initialized.

Does that make sense? I haven't seen this type of service init in our code, but I am thinking why it's a bad idea and can't think of anything (apart from the manual digest, but that's not too terrible). On the plus side, it is easier to integrate into controllers (will be done in one of the jiras later).
What do you think?

vrathee:
Reviewed!

PR #537, opened 2016-05-31
Reverting PCM-1857

jtrantin:
I am disabling the functionality as discussed in PCM-1857. Just commenting the code because we will want to re-enable it again when a full investigation on what fields should be sent is done.


renujhamtani:
Loks good!

PR #547, opened 2016-06-10
Fixing the udsjs issues

engineersamuel:
@jtrantin I'm not too sure I did much here beyond ensure the dist/uds.js was being used and to upgrade jsuri, but it does work in local dev without errors.  Can you review this and see what we need to do next if anything?

I'm still not that familiar with the build.  It seems strange to declare the udsjs in the chrome.html and then also load it with the loader.js.  I commented out udsjs in the chrome.html file.  I'm not sure what the relevance of the ascension.html and index.html are or if they are even used.

### DIFF
--- a/app/chromed.html
+++ b/app/chromed.html
@@ -46,7 +46,7 @@
 
     <script>
     window.deps = [];
-    window.deps.push("/bower_components/jsUri/Uri.min.js");
+    window.deps.push("/bower_components/jsUri/Uri.js");
     window.deps.push("/localFile.js");
 
     window.deps.push("/bower_components/bootstrap/dist/js/bootstrap.js");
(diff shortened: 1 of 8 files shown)


engineersamuel:
@vrathee @gandhikeyur since Jonas is out, can you guys review this?  Jonas was having issues with the PCM build with the new udsjs, I did see some issues, and just did some general clean up.  This PR should fix any problems, but I don't work in the PCM that much right now, if you two could review this, it'd be good to get it merged if necessary and not wait until @jtrantin get's back.

app/chromed.html
```diff
@@ -64,6 +64,7 @@
     window.deps.push("/bower_components/bluebird/js/browser/bluebird.js");
     window.deps.push("/bower_components/angular-md5/angular-md5.js");
     window.deps.push("/bower_components/pegjs/peg-0.9.0.js");
+//    window.deps.push("/bower_components/udsjs/dist/uds.js");
```
vrathee: @engineersamuel  Aren't we uncommenting the uds.js lines throughout PCM?

engineersamuel: The uds.js is already added in the loader.js, it doesn't seem to be needed here as that would be a duplicate, if I understand correct.  What do you think?

vrathee: @engineersamuel  We use loader.js only for local development.. for itos we have is configured separately in the index.html file itself. 
But since we anyways don't use udsjs in PCM so it won't create any problem :)

vrathee:
Reviewed!

gandhikeyur:
I think the changes look good. Reviewed!!

engineersamuel:
@vrathee @gandhikeyur I unmarked as WIP, feel free to merge!

PR #549, opened 2016-06-23
PCM-2901

engineersamuel:
Moving the PCM to webpack.


engineersamuel:
@vrathee @gandhikeyur @jtrantin please begin reviewing.  I merged to upstream today, resolved conflicts, however the travis build isn't working, and it isn't exactly clear why at the moment.   Considering the number of changes I've made, I expect there to be potential regressions, so please review very closely.

jtrantin:
@engineersamuel The build failed executing `npm install`. Looks like a npm dependency issue -- https://travis-ci.org/redhataccess/redhat_access_angular_ui/builds/139856346 starting at line 270. Will dive into the code now..

jtrantin:
After going through the code and trying it for myself I have the following:
 - Not sure why Travis is failing, `npm install` is working fine for me. Maybe it has bad node/npm version?
 - uibModal in new.js is not converted, lines 4 and 342 causing /cases/new to fail
 - Support Cases heading is missing, see screens below
 - CSS is not applied correctly see screenshots below (I think we faced similar issue in Ascension too)
 - Group dropdown contains bad characters

However, when I fixed the modal I was able to create a case and navigate without problems, so I'd say this is looking very good!
Screens: https://goo.gl/photos/ACLR84uXoZ3WdjWo9

renujhamtani:
@engineersamuel We are good to merge this. Could you please rebase this to the latest and solve the problems which jonas is facing? Once you have done this, jonas, keyur and vikas can review this and once done, will start with smoke testing.

engineersamuel:
@jtrantin @vrathee @gandhikeyur  Do you guys see something I'm missing?  I tried the build again in travis with no luck.  It appears to be failing on grunt-autoprefixer, which is not in the package.json, it use to be, now it just has the 'autoprefixer' package which has no dep on 'grunt-autoprefixer'.  I wonder if there is something cached that is interfering.  It seems unlikely, but I'm trying to think of anything.

I installed nvm on my local box and installed node .10 to make sure that wasn't the prob, and it installed fine without issue.  Definitely no question I'm using the right node version.  The npm version may be different, I'm on 2.15.1, but I can't imagine that is a problem as native packages would be compiled against node, the npm wouldn't matter.

Definitely open to suggestions for what I can try next.

jtrantin:
Strange, I know you say the node version isn't an issue, still it seems rather old. I am using LTS version v4.4.4 and I know I've been having problems with older versions. Check https://nodejs.org/en/ for more info about versions.
Also, it seems there are some leftovers, because it is installing that dependency. Maybe running install with --dry-run would produce some insights on why it is installed.

engineersamuel:
@jtrantin right, so I was running a later version locally, but saw that travis was running an earlier version, so I used nvm to try to match travis as closely as possible.  I guess you mean add --dry-run to the travis scripts?  Heck, maybe we should just upgrade the PCM gears to a later version, but I feel like that opens another can of worms right now.  I'm hoping not to touch was isn't broken in that regard unless we absolutely have to at this exact point in time.

jtrantin:
Yes, I suggest adding npm install --dry-run to travis build scripts as that might help diagnose the problem. 
I think we will need to update PCM gears with later node version, because, at least for me, even webpackized Ascension didn't work before I updated the node version.

renujhamtani:
@engineersamuel  The travis build looks clean now. But i guess there are yet merge conflicts. Can you please resolve those.  Is there anything else besides merge conflicts which we need to resolve?? 
@gandhikeyur @vrathee Please review the changes.

engineersamuel:
@renujhamtani @jtrantin @vrathee @gandhikeyur The build is good to go now, we can merge, I know we'll need further changes in the gears and potentially for gs4 as well, but I suspect we'll have to do those after the merge.

jtrantin:
Some of the things I mentioned earlier:
 - Support Cases header is missing
 - Portal CSS is not applied correctly, now it looks more like stock bootstrap.
 - Group dropdown contains bad characters

See my previous comments for screens.

Other issues I just noticed:
 - The manage bookmarked accounts button on the list page (the one with star and users icons) acts funny. It seems the popup covers most of the button and that makes it impossible to click it.
 - Accessing case 01323723 gives an error `Cannot read property 'number' of undefined`. Note you have to navigate to the case from the app, if you go directly to the case page it doesn't occur; just go to /case/list and replace list with the case number.
 - The advanced search (at #/case/search) doesn't work, not sure about this but I think it worked the last time I tried it. We need to require PEG.js in solrGrammarService.

engineersamuel:
@jtrantin Yep, going to work on all that this morn.

engineersamuel:
@jtrantin regarding the header not being shown, it is definitely in the DOM: http://cloud.samuelmendenhall.com/1O36212N2R3P  it just appears there are no parts of the header to show as they mainly include error or warning situations or am I missing something?

engineersamuel:
@jtrantin 


Support Cases header is missing (in /case/list and /case/new)

UNKNOWN: It is definitely in the DOM, can you show me why believe it is missing?

Portal CSS is not applied correctly, now it looks more like stock bootstrap.

UNKNOWN: It appears to look fine to me, not necessarily sure what you are referring to.

Group dropdown contains bad characters

UNKNOWN: I saw this when I was doing the original jira, and it was fixed when I created the initial PR, can't reproduce this, send screens in exactly where you are looking at so I can try to reproduce.

The manage bookmarked accounts button on the list page (the one with star and users icons) acts funny. It seems the popup covers most of the button and that makes it impossible to click it.

DONE: Had to attach tooltip to body instead of element

Accessing case 01323723 gives an error Cannot read property 'number' of undefined. Note you have to navigate to the case from the app, if you go directly to the case page it doesn't occur; just go to /case/list and replace list with the case number. The page seems to be working fine, apart from the recommendations, which are gone so it's maybe related to that.

UNKNOWN: I don't doubt this is a bug, but I can't reproduce it.  I don't know why going directly to the case wouldn't reproduce it as that is what happens when you click on a case in the list regardless:  `a(href='#/case/{{case.case_number}}')`  The ui router should be invoked in the same matter whether you click on a case in the list or you manipulate the url.  Any other ideas on reproducing this?

The advanced search (at #/case/search) doesn't work, not sure about this but I think it worked the last time I tried it. We need to require PEG.js in solrGrammarService.
When I load /case/list the list of cases seems to be duplicated. https://goo.gl/photos/JPT8ne3ncpbbtmBTA

DONE: Fixed, required in the pegjs.

jtrantin:
Support Cases header is missing (in /case/list and /case/new) -> I mean the Support Cases heading is missing, check the album for details -- https://goo.gl/photos/ACLR84uXoZ3WdjWo9. You can see that on the list page, above the filtering section, there is no Support Cases heading in your version, but it is in QA. This also applies to case/new and other places that use the heading from the header component (e.g. advanced case search is not using that).

Portal CSS - probably a caching issue, now all looks good.

Group contains bad characters. -- probably caching issue, now I see the proper separators. However, the separators are selectable,  as can be seen here -- https://goo.gl/photos/14kChmv4CfzG6yMy9 (check the comments on the images for details).

Otherwise looking good, advanced case search fully functional!

gandhikeyur:
Did not checkout & run it locally, but the code changes look good. I was mostly scrolling through the changed files as the number of files changed is massive.
Should be good to merge if the issue's mentioned by jonas are fixed.

engineersamuel:
@jtrantin @vrathee @gandhikeyur Pertaining to the header, I know why this is happening, but I don't know the solution.  Maybe one of you will have an idea.  Look at this screen: http://cloud.samuelmendenhall.com/3V0U1v122Z1Z

The header contains the following which should display the title: `div(ng-hide='page ===""', rha-titletemplate='', page="{{page}}")`

That directive is:

```
export default function () {
    return {
        restrict: 'AE',
        scope: {page: '@'},
        template: require('../views/title.jade'),
        controller: 'TitleViewCtrl'
    };
}
```

Simple enough, it should have a scope with the page variable (which it does) and with the controller variables and functions (which it does *not*) https://github.com/redhataccess/redhat_access_angular_ui_common/blob/master/app/common/controllers/titleView.js

I'm not quite sure why the TitleViewCtrl vars and getPageTitle are not in the scope.  I'm assuming this is some behavior of Angular isolated scopes I'm not familiar with.

If I look one parent up, the scope is fine and includes the header.js controller info as you'd expect: http://cloud.samuelmendenhall.com/462k1R2B2a1u

engineersamuel:
@jtrantin I swear I about pulled my hair out on the disabled options one.  Long story short, there is a lot of misinformation on google about how to disable options with angular-chosen.  Most of it pertains to Angular < 1.4.  I finally found one little nugget in stackoverflow that says Angular > 1.4 to simply `<select ng-options="c.name disable when c.shade == 'dark' 
group by c.shade for c in colors">`  And it works like a charm.

engineersamuel:
@jtrantin Could you verify if the https://qa.foo.redhat.com:1337/logout#/case/new should or should not default to selecting the sep option, right now it is since with a new case there is no kase product to match on so it matches on the empty or null option which is the separator.  I believe this is correct behavior but please verify.  (I think a 'Select Option' initial option would be better but I'll defer to you)

jtrantin:
@engineersamuel  I found the culprit for the headings, check titleView.js in common module, there is `function` instead of `contructor`. I tested it and it works with `constructor`.

The `/case/new` page works just fine, the only thing is separators are selectable in product selectbox. You might want to fix this throughout the app.

engineersamuel:
@jtrantin Ok, we have a new release of the commons with that function -> constructor, so we are looking good there.   I also fixed the `/case/new` product options.  I don't believe there are any other `options-disabled` directives in use.

@jtrantin @vrathee @renujhamtani @gandhikeyur  I think we are incredibly close, if not finally ready, to be able to merge. 

PhantomJS 2.1.1 (Mac OS X 0.0.0): Executed 207 of 213 (skipped 6) SUCCESS (1.543 secs / 1.414 secs)

PR #550, opened 2016-06-24
Pcm 2955

gandhikeyur:
@vrathee @renujhamtani Please review

### DIFF
--- a/app/ascension.html
+++ b/app/ascension.html
@@ -52,7 +52,7 @@
     <![endif]-->
     <!-- Manually add uri and angular treeview due to bower issues -->
 
-    <script src="bower_components/jsUri/Uri.js"></script>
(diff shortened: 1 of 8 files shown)


gandhikeyur:
Merging this..

renujhamtani:
We will check in CI

@engineersamuel  Just an FYI. We are reverting the PR which  u created for udsjs change as PCM CI is giving error. We will talk in scrum.This works locally fine but on CI, it gives error. This was the PR  https://github.com/redhataccess/redhat_access_angular_ui/pull/547

engineersamuel:
Ok, I wish I could have spent 5 minutes testing this in CI, I bet I could have fixed it.   The only issue I ever saw was the un-transpiled uds.js was used instead of the transpiled one, when I tested locally it was a really easy fix.

PR #556, opened 2016-07-22
PCM-3074: Fixing the product select dropdown

engineersamuel:
The product dropdown was defaulting to the separator and the sep wasn't disabled, now it no longer will.

### DIFF
--- a/app/cases/services/productsService.js
+++ b/app/cases/services/productsService.js
@@ -74,7 +74,7 @@ export default class ProductsService {
 
                 const sep = '────────────────────────────────────────';
                 if (productOptions.length > 0) {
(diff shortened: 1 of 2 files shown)


app/cases/services/productsService.js
```diff
@@ -74,7 +74,7 @@ export default class ProductsService {
 
                 const sep = '────────────────────────────────────────';
                 if (productOptions.length > 0) {
-                    productOptions.push({ isDisabled: true, name: sep });
+                    productOptions.push({ isDisabled: true, name: sep, code: '' });
```
engineersamuel: This appears necessary otherwise the default is null which will be what the ng-model is and it will be selected.

jtrantin:
Reviewed.

PR #559, opened 2016-08-09
PCM-3144: Reverting to the version of ng table prior to the webpack changes

engineersamuel:
https://projects.engineering.redhat.com/browse/PCM-3144

With moving to npm I had used the latest version of ng-table which has completely change how it does things.  I didn't see the 'Manage case groups' as it was at the end of the dropdown so I never exposed the bug here until qa pointed it out.  Reverting to the the 0.3.3 version fixes the problem with no additional code changes.


engineersamuel:
@jtrantin @vrathee @gandhikeyur Please review.

renujhamtani:
Reviewed! Looks good!

PR #561, opened 2016-08-10
PCM-3140: Fixed solution linking/pinning

jtrantin:

https://projects.engineering.redhat.com/browse/PCM-3140

Strangely this code was in the app for quite a long time, so maybe strata did some changes that introduced the pinning problems. Note that before applying this fix, the pinning worked sometimes, but sometimes it didn't. With this fix, the pinning works much better, I haven't been able to reproduce any of the problems.


renujhamtani:
Looks good!

PR #562, opened 2016-08-15
PCM-3167: IE11- UI is looking weird for PCM

engineersamuel:
https://projects.engineering.redhat.com/browse/PCM-3167

Reverting to angular-chosen-localytics v1.0.6 as the latest version has bugs with IE11


engineersamuel:
@jtrantin @gandhikeyur @vrathee Please review.  I was able to get about 90% of the styling working with the latest chosen version, but there were a few things that still didn't look right.  There are apparently some issues with the latest chosen version and IE11+ so I'm going to just play it safe and use the last known working version.  There was still one IE11 css hack I had to make, beyond that, there are difference between IE11 and Chrome, but those are visual only, no misalignment or hidden text anymore.

app/assets/sass/vendor/_chosen.scss
```diff
@@ -5,6 +5,22 @@ $chosen-button-bg-color: $brand-gray;
 	width: 100%;
 }
 
+// http://stackoverflow.com/questions/20541306/how-to-write-a-css-hack-for-ie-11
```
jtrantin: Do we really need to include this commented-out fix? I understand it is probably for future reference, but I don't see us updating the package until most of these issues are resolved.

app/assets/sass/vendor/_chosen.scss
```diff
@@ -106,10 +122,20 @@ $chosen-button-bg-color: $brand-gray;
 			border: none;
 
 			span { color: #f0f0f0; }
+            // Force the right background in ie11
+            span {
+                @media all and (-ms-high-contrast: none), (-ms-high-contrast: active) {
+                    color: $chosen-button-bg-color !important;
```
jtrantin: Is this correct? The comment says this should set background color, as does the SASS variable, but you are setting foreground `color`.

vrathee:
Reviewed! Apart from Jonas's comment on foreground color.

app/assets/sass/vendor/_chosen.scss
engineersamuel: Yes this is intended.  It looks odd because it is an IE11 hack.  Even with the old chosen version back in there is an issue with the text coloring only, this fixes that in IE11+.  The comment is wrong, I'll fix that so it isn't confusing.

app/assets/sass/vendor/_chosen.scss
engineersamuel: We can remove it as I already have the hack in the code so we can just copy and paste it when needed.

jtrantin:
Reviewed. Once the build is green we can merge.

engineersamuel:
@jtrantin @renujhamtani I had to edit the travis build to remove node_modules to fix this build.  I've reverted those changes so all is good now.  It always worked locally, which got me thinking the node_modules were being aggressively cached by travisci, which is what was happening, even with the npm cache clean.  I've reverted those changes but kept it documented in a comment in case we hit this in the future.

I believe we are ready to merge this now.

PR #563, opened 2016-08-17
Pcm 1529 -  Sending new event on case creation page interaction and sending package json version

vrathee:
@engineersamuel @jtrantin Please review.

https://projects.engineering.redhat.com/browse/PCM-1529
https://projects.engineering.redhat.com/browse/PCM-3163


app/main.module.js
```diff
@@ -39,6 +39,8 @@ if (ENVIRONMENT === 'test') {
     require('angular-mocks');
 } else {
     require('./assets/sass/main.scss');
+    const packageJson = require('./../package.json');
+    window.pcmVersion = packageJson.version;
```
vrathee: For - https://projects.engineering.redhat.com/browse/PCM-3163

Use version mentioned in package.json while sending recommendations with case creation

engineersamuel:
Reviewed, afaict everything looks good!

renujhamtani:
@sshumake This is the PR for https://projects.engineering.redhat.com/browse/PCM-1529. It would  be good if you can also loook at it. I am merging it now.

app/main.module.js
jtrantin: Seems a bit strange, to require it and bind it to window. Why don't we just require it wherever we actually need it (`caseService.js` in this case). Since this is already merged, we can let it be as is.

jtrantin:
Looks good.

PR #565, opened 2016-08-22
Adding check for new case url

vrathee:
@renujhamtani @gandhikeyur Please review.

The product and version components calling this function are used on edit page as well. This check will make sure we don't fire this event from any other page.


vrathee:
https://projects.engineering.redhat.com/browse/PCM-3213

gandhikeyur:
Reviewed

PR #566, opened 2016-08-26
PCM-3025: Added query to URL

jtrantin:

https://projects.engineering.redhat.com/browse/PCM-3025

I also added a fix for the same problem as in PCM-2943, and some more small fixes usually related to the advanced search.

### DIFF
--- a/app/assets/sass/elements/_base.scss
+++ b/app/assets/sass/elements/_base.scss
@@ -153,3 +153,10 @@ a { color: $link-color; }
 		margin-right: 8px;
 	}
 }
+
+.spinner.ng-animate {
+    transition:0s none;
+    -webkit-transition:0s none;
+    animation: 0s none;
+    -webkit-animation: 0s none;
+}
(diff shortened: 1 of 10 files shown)


app/assets/sass/elements/_base.scss
```diff
@@ -152,4 +152,11 @@ a { color: $link-color; }
 		color: #afafaf;
 		margin-right: 8px;
 	}
-}
\ No newline at end of file
+}
+
+.spinner.ng-animate {
```
jtrantin: Fixes 'hanging' spinners, as in PCM-2943.

renujhamtani:
We would merge this after the PCM prod push.

jtrantin:
OK. If I manage to finish PCM-3024 by then, we can do it in one PR, since I rewrote some of the code introduced here.

jtrantin:
Since most of the code here is rewritten/removed in the next issue I am working on, I will submit the other one which will include both JIRAs (PCM-3025 + PCM-3024). Let me know if that is a problem.

PR #567, opened 2016-09-12
PCM-3309: Removing ngAnimate module

jtrantin:
@vrathee @gandhikeyur @engineersamuel @renujhamtani 

https://projects.engineering.redhat.com/browse/PCM-3309

How do you feel we remove the ngAnimate module? AFAIK it is not used anywhere in the project and it only causes delays when hiding elements. I tested this change, and removing it doesn't seem to have any negative effects.


engineersamuel:
Ack, unless I'm forgetting something, should be good to go.

gandhikeyur:
Reviewed

PR #569, opened 2016-09-12
PCM-3311: Implementing pagination

engineersamuel:
https://projects.engineering.redhat.com/browse/PCM-3311

Migrating from ng-infinite-scroll to pagination


engineersamuel:
@jtrantin @gandhikeyur @vrathee Please review!

engineersamuel:
@jtrantin @gandhikeyur @vrathee Please review!

app/cases/filters/startFrom.js
```diff
@@ -0,0 +1,8 @@
+'use strict';
+
+export default function () {
+    return function(input, start) {
+        start = +start; //parse to int
+        return input.slice(start);
```
jtrantin: Looking at this directive, does it only skip first X elements from the input? I think you can do the same with limitTo, just specify it's second parameter, see https://docs.angularjs.org/api/ng/filter/limitTo.

app/cases/views/advancedSearch.jade
```diff
@@ -11,13 +11,13 @@ div(ng-if='securityService.loginStatus.isLoggedIn')
...
             .alert.alert-warning No cases found.
-        div(ng-show="AdvancedCaseSearchService.query != null && AdvancedCaseSearchService.cases.length > 0 && AdvancedCaseSearchService.allRecordsCount != null")
-            span(translate) Showing {{AdvancedCaseSearchService.cases.length}} out of {{AdvancedCaseSearchService.allRecordsCount}} search results
+        div(ng-show="AdvancedCaseSearchService.query != null && AdvancedCaseSearchService.cases.length > 0 && AdvancedCaseSearchService.totalCases != null")
+            span(translate) Showing {{AdvancedCaseSearchService.getCasesStart()}} to {{AdvancedCaseSearchService.getCasesEnd()}} of {{AdvancedCaseSearchService.totalCases}} search results
 
-    #case-list.table-list.case-list(ng-hide="RHAUtils.isEmpty(AdvancedCaseSearchService.cases)", infinite-scroll="bottomScrolled()")
-        .list-item(ng-repeat='case in AdvancedCaseSearchService.cases')
+    #case-list.table-list.case-list
+        .list-item(ng-repeat='case in AdvancedCaseSearchService.cases | startFrom:AdvancedCaseSearchService.currentPage*AdvancedCaseSearchService.pageSize | limitTo:AdvancedCaseSearchService.pageSize track by $index')
```
jtrantin: Can `getCasesStart` be used here instead of `AdvancedCaseSearchService.currentPage*AdvancedCaseSearchService.pageSize`? Seems like the expressions are identical.

app/cases/views/advancedSearch.jade
```diff
@@ -50,3 +50,11 @@ div(ng-if='securityService.loginStatus.isLoggedIn')
         p.center
             span.spinner.spinner-inline
             span {{'Loading cases...'|translate}}
+    //{{AdvancedCaseSearchService.currentPage+1}}/{{AdvancedCaseSearchService.numberOfPages()}}
```
jtrantin: This commented out code can be removed.

jtrantin:
Some things I noticed while testing locally. 

- In Advanced Search, the pages are reloaded with each navigation, while in case/list the results are cached. I think we should also cache the results in advanced search too, to make navigation between already loaded pages faster.
- When you go to a case from advanced search and then hit Back you can only page within already loaded pages, when you navigate to a new page, instead of loading it it shows a blank screen. Note the search query doesn't have to be persisted, but the case results should stay (check the current behavior in QA).
- Sometimes in case/list, when going to a case and hitting Back, the previously displayed cases disappeared, and only appeared when I switched a couple of pages. This only seems to happen for later pages (didn't happen for first page, try the last page).

app/cases/filters/startFrom.js
engineersamuel: There was a stackoverflow on pagination with ng-repeat, copied it from that: http://stackoverflow.com/questions/11581209/pagination-on-a-list-using-ng-repeat

I see from the docs you posted though that limitTo has a start option, so yes, I can give that a try.  Either way should work.

app/cases/views/advancedSearch.jade
engineersamuel: Thought I caught that everywhere and replaced it with the getCasesStart|End, will change.

app/cases/views/advancedSearch.jade
engineersamuel: I left it in as a reference for showing page/totalpages, but I'll remove it if you wish.

app/cases/filters/startFrom.js
engineersamuel: Just tried limitTo:limit:start out and it works great, will remove the startFrom filter.

engineersamuel:
@jtrantin 

1. Done.

For 2. and 3. I am so on the verge of using Redux here, but I really don't want to spend any more time on this.  I've spend probably 3-4 hours today attempting to debug especially 2.  The problem is the isolated scope.  When doing an advanced search, clicking a case, then clicking back the scope is blown away.  Even when I tried binding the query to $parent.AdvancedCaseSearchService.query it still blows that away.  Even when I set the AdvancedCaseSearchService.query in the solrQueryInput.js controller when the query is updated, it is still nuked when I go back.

I've tried the following options:
* solr-query with and without $parent
* ng-if and with ng-show, no difference, though ng-show does require changes, but no difference in the scope issue
* Tried with referencing AdvancedCaseSearchService for the query and queryParsed and just as scope vars, no difference

When going to a case then back the query is never retained in the input element, and the queryParsed is always false.  This means subsequent Next buttons will never behave right since there is no query set, that's the problem.

I believe this is the same issue for /case/list.

I welcome any suggestions here, but I'm tired of fighting angular.  I hope this is just something simple that I'm not setting.  The quick and dirty solution is to just open the cases externally on click.  The better solution, imo, would be to use redux and use top down state which means that when navigating back to the search page the global state would be read and be identical.  I'm not sure how the solr-query.

I fully expect you to *not* like target="_blank" as a solution, but give me some ideas or alternatives, as I'm fresh out besides redux.  The target blank would also need changes as the styling is off: http://cloud.samuelmendenhall.com/0X3v0o3z0z02

engineersamuel:
@jtrantin  Thinking some more, I think we have 5 options:

1. Use target _blank for case links (Cheap and easy, but a change in behavior that may not be welcome)
2. Figure out the scoping issue (Probably a really simple change, but I can't find the right incantation)
3. Use Redux (Would be awesome, but time consuming, and do we want to take the time here for this in September?)
4. Bookmarkable URLs (Now this an option I really really like, and something we should have regardless.   This would solve the problem, but potentially take some time to implement)
5. Redux + bookmarkable URLs (The ultimate solution, but the most time consuming one.

So unless you have an idea for 2. and 1. is probably out of the question, then 4. may be the best option right now.    Let me know what you think.

jtrantin:
In case/list I have tracked the problem down to `SearchCaseService.clearPagination()` in `doSearch` in `list.js` controller. The problem is, that after returning back to /case/list, the init function is called, which triggers `doSearch` and that clears pagination, which clears all the cases. This basically erases all the pages before the current one. Looking into it more now, I wonder if that clear pagination is really needed or not.

jtrantin:
I take that back.. the pagination clearing is needed, however, we can add a further check and only trigger `doSearch` on controller init if there are no cases already. So instead of `$scope.doSearch();` we'd have `if(RHAUtils.isEmpty(SearchCaseService.cases)) $scope.doSearch();`. Tested that works well.. now I'll look into advanced search

gandhikeyur:
Reviewed. Merging this..

PR #571, opened 2016-09-19
Chosen CSS fixes

jtrantin:
@renujhamtani @vrathee The fix for missing X in account select, please review.

I just removed chosen.css that was included into the SASS, we already have chosen CSS provided by chromed themes. Also added top margin to make it centered.


vrathee:
Reviewed!

renujhamtani:
Reviewed!

PR #573, opened 2016-09-20
PCM-3350: Fix for cases not being fetched

jtrantin:
@renujhamtani @vrathee @gandhikeyur Please review, this fixes the case-loading blocker

https://projects.engineering.redhat.com/browse/PCM-3350

What happens in PCM is that if the request for cases (/case/filter with SSO in the payload) fails, PCM runs another query, this time without the SSO. The one line in `searchCaseService.js` prevented the other request form being executed. Not sure why two requests need to be executed, or why the first one fails. Strata team should still investigate why the first requests fails with 404.

### DIFF
--- a/app/cases/controllers/bookmarkedAccountSelect.js
+++ b/app/cases/controllers/bookmarkedAccountSelect.js
@@ -27,8 +27,10 @@ export default class BookmarkedAccountSelect {
             }
         };
 
-        $scope.$watch('selectedAccount', function () {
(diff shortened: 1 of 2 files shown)


vrathee:
Reviewed! It will also fix the console error ryt?

jtrantin:
As discussed on IRC, it should fix the console error we were seeing, though I would verify that once on QA.

PR #574, opened 2016-09-21
Fixing PCM-3359

vrathee:
@jtrantin @gandhikeyur Please review.

I also saw that ascension logo was visible while running PCM for some seconds while loading.. So fixed that as well..

Regarding the current fix. I feel it should have been in the old PCM as well but don't know why we never got it.


renujhamtani:
Reviewed!

PR #577, opened 2016-09-21
PCM-3365: Fixing trim of undefined errors

jtrantin:

https://projects.engineering.redhat.com/browse/PCM-3365

This error kept popping up for Spenser and other users. Since there is just few occurrences of trim I checked them all and fixed those that could be problem. Verified on new case view, where even I was getting the error.


gandhikeyur:
Reviewed

gandhikeyur:
Merging this for the QA to test early as we want to do a prod push today

PR #579, opened 2016-09-22
PCM-2745: Added DNF message to account search

jtrantin:
@vrathee @gandhikeyur @engineersamuel  Please review.

https://projects.engineering.redhat.com/browse/PCM-2745

### DIFF
--- a/app/cases/controllers/accountSearch.js
+++ b/app/cases/controllers/accountSearch.js
@@ -31,15 +31,15 @@ export default class AccountSearch {
                         $scope.account = AccountService.accounts[number];
                         $scope.valid = true;
                         $scope.invalid = false;
+                        $scope.loading = false;
                     }
-                    $scope.loading = false;
                 } else {
                     if (number == $scope.accountQuery.trim()) {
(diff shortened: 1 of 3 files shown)


app/cases/controllers/accountSearch.js
```diff
@@ -28,22 +27,25 @@ export default class AccountSearch {
...
                 } else {
-                    $scope.account = null;
-                    $scope.invalid = true;
-                    $scope.valid = false;
+                    if (number == $scope.accountQuery.trim()) {
+                        $scope.account = null;
+                        $scope.invalid = true;
+                        $scope.valid = false;
+                    }
                     $scope.loading = false;
```
engineersamuel: Very minor, but both the if and else set `$scope.loading = false` which means you can move it outside the if/else as it will be set to false regardless or remain false.

jtrantin: Actually, putting the `loading = false` into the ifs is even better, now the spinners are shown correctly even if multiple overlapping requests are fired.

PR #580, opened 2016-10-05
PCM-3460: Setting redhat_client for Strata

jtrantin:

https://projects.engineering.redhat.com/browse/PCM-3460

This piece was originally present in the index.html in CI gear repo, and it got deleted when we switched to webpack. Confirmed on QA env it sets the right Case Origin for new cases.

### DIFF
--- a/app/main.module.js
+++ b/app/main.module.js
@@ -29,19 +29,19 @@ require('chosen/chosen.jquery.js');
 window.moment = require('moment');
 require('moment-timezone');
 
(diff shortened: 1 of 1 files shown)


vrathee:
Reviewed!

jtrantin:
Just pushed commit to fix tests. We don't want to load package.json in test environment, so I just moved the code to a near if statement.

gandhikeyur:
Reviewed

PR #582, opened 2016-10-12
PCM-3339: Fixed breadcrumbs for Search and Groups

jtrantin:

@renujhamtani While fixing this I noticed there are two routes we probably don't use anywhere. Both `editGroup` and `defaultGroup` routes seem to have been replaced by the manage groups interface. Should I remove those routes?


vrathee:
Reviewed!

renujhamtani:
@jtrantin please remove those 2 routes which are not used.

PR #587, opened 2016-10-17
PCM-3334 Fix shift position issue when submit request management escalation

kunyan:


kunyan:
https://projects.engineering.redhat.com/secure/attachment/21884/Screencast%20from%2010-17-2016.webm

PR #588, opened 2016-10-17
PCM-3331 Fix pag number incorrect issue

kunyan:
1. Page will start from 1
2. Fix page number is 0 error when do filter
3. Fix page number incorrect on advancedSearch

### DIFF
--- a/app/cases/controllers/advancedSearchController.js
+++ b/app/cases/controllers/advancedSearchController.js
@@ -65,6 +65,7 @@ export default class AdvancedSearchController {
                 const query = SOLRGrammarService.parseSafe($scope.solrQuery);
                 if(query != null ) {
                     $scope.$broadcast(CASE_EVENTS.advancedSearchSubmitted);
+                    AdvancedCaseSearchService.currentPage = 1;
                     AdvancedCaseSearchService.performSearch(query, CaseService.filterSelect);
                 }
             }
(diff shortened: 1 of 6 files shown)


gandhikeyur:
@kunyan Going ahead, please tag atleast two developers from our team to review the PRs that you send.
@jtrantin Could you please review this.

jtrantin:
@kunyan 
Since we are showing "1 to 20", wouldn't it make sense to show "21 to 40" and not "20 to 40". Also, we don't show `currentPage` anywhere in the app, so it doesn't have to start with 1, changing it only makes the code changes more complicated, without any actual benefit.
I assumed this would be just a simple fix of adding +1 to the first number in the "Showing x to y" message.

kunyan:
@jtrantin 

even we didn't show currentPage in anywhere, but I think we should correct it from 0 to 1. the old way is in order to set offset to strata api.  make the currentPage to be current page would be better.

I'll change getCaseStart to 

```javascript
this.getCasesStart = () => {
    return (this.currentPage - 1) * this.pageSize + 1;
}
```

jtrantin:
Not sure why starting page from 1 is better than starting from 0, if we are not showing this information to the user. It just seems these changes are overly complicated, since PCM-3331 could be fixed with 4 characters (just adding +1 to the jades, st. like `{{AdvancedCaseSearchService.getCasesStart()+1}}`). 
I personally find it easier to keep the "strata" offset and page number and then modify it when showing to the user, but I guess that is just user preference. 

But OK, if you fix the `getCaseStart` as you mentioned we can merge this.

kunyan:
@jtrantin  getCaseStart has been fixed.

PR #590, opened 2016-10-19
PCM-3330 Make the invalid group error message more friendly

kunyan:


kunyan:
@gandhikeyur @jtrantin pls review it

app/i18n/template.pot
```diff
@@ -955,6 +955,10 @@ msgstr ""
 msgid "Product Version"
 msgstr ""
 
+#: app/cases/controller/detailsSection.js
```
jtrantin: I think you don't need to touch these files, as they are generated somehow. @vrathee @gandhikeyur Can you confirm these changes are ok an will not break anything?

renujhamtani:
@kunyan we dont touch the template.pot files. You can remove those changes from this PR. After that, I will merge this.

renujhamtani:
looks good now.

PR #591, opened 2016-10-19
PCM-3257: Added session.js

jtrantin:
@vrathee @gandhikeyur @renujhamtani Please review.

https://projects.engineering.redhat.com/browse/PCM-3257

I just wrapped the current auth logic in session.onInit. The current auth logic is good and will keep working even with keycloak, we just need to wait until session.js does its thing.


vrathee:
Reviewed!

PR #595, opened 2016-11-04
PCM-3609: Fixing public search link

jtrantin:

https://projects.engineering.redhat.com/browse/PCM-3609

The `absolute: true` generated absolute URL without the proper path. To fix this, I removed generate relative url, and let browser handle the conversion to absolute. Opening in a new tab and copying with right-click still works fine.


vrathee:
Reviewed!

PR #597, opened 2016-11-10
Fixing translations

jtrantin:
@vrathee @gandhikeyur @renujhamtani Please review.

https://projects.engineering.redhat.com/browse/PCM-3632

From my limited knowledge of translations, I put together a new way to generate translation templates. Instead of grunt I use angular-gettext-cli, and the generation is done in two parts:

- scraping the dist .js file
- srcaping the view files

Since we no longer generate .html, and since the angular-gettext-cli doesn't understand .jade, I had to add a compilation from jade to html as well. Please, think about this PR and ask questions if something doesn't feel right.

To build the templates you need to install **angular-gettext-cli** with `npm install -g angular-gettext-cli`, **jade** with  `npm install -g jade` and **msgcat** (that was bundled with my fedora already).

Then the procedure is the following:

1. Do a regular build with `npm run build`
2. Generate translate templates with `npm run generate-translates`

### DIFF
--- a/app/cases/views/discussionSection.jade
+++ b/app/cases/views/discussionSection.jade
@@ -131,10 +131,10 @@ ul#tab_list.nav.nav-tabs(role="tablist")
       .row.row-short
         .col-sm-7
           span#notesNotice.uploadNotice
-            span.progressCount(translate) {{noteCharactersLeft}} {{'characters left.'|translate}}
+            span.progressCount() {{noteCharactersLeft}} {{'characters left.'|translate}}
         .col-sm-5
           .pull-right
-            button.btn.btn-app(ng-click="updateNotes()", ng-disabled='CaseService.updatingCase || !notesForm.$dirty',translate='') {{'Update'|translate}}
+            button.btn.btn-app(ng-click="updateNotes()", ng-disabled='CaseService.updatingCase || !notesForm.$dirty') {{'Update'|translate}}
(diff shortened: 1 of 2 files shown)


jtrantin:
Note the generated translations will look different (different order as well as keys), but I went through the new vs existing keys, and it looked like nothing is missing.

Also, since we are building the keys off the dist/pcm.js we get translates for the React files automatically in hybrid architecture. So, we can keep using `gettextCatalog` in React too. I checked existing React solutions, but there doesn't seem to be anything substantially better/different, so I don't think it's worth the hassle to move to some react-specific solution for translates.

app/cases/views/discussionSection.jade
```diff
@@ -131,10 +131,10 @@ ul#tab_list.nav.nav-tabs(role="tablist")
       .row.row-short
         .col-sm-7
           span#notesNotice.uploadNotice
-            span.progressCount(translate) {{noteCharactersLeft}} {{'characters left.'|translate}}
+            span.progressCount() {{noteCharactersLeft}} {{'characters left.'|translate}}
```
jtrantin: Removing double translate, the same for the change below.

gandhikeyur:
Changes look good.. Kudos for fixing this super quick!!

vrathee:
@jtrantin  Looks great :) . Maybe we can also have a task to clear genrated html files for the views directory.

jtrantin:
Good idea Vikas! Added a script that removes all files generated during the process. I checked all files before and after generating templates and got the same set of files, so we should be good :).

vrathee:
Re-Reviewed :)

PR #598, opened 2016-11-15
PCM-2957 + PCM-3162: Attachment uploading improvements

jtrantin:
@vrathee @gandhikeyur @engineersamuel Please review.

https://projects.engineering.redhat.com/browse/PCM-2957
https://projects.engineering.redhat.com/browse/PCM-3162

This consists of two separate features. First is to show more human-friendly error if user tries to upload file he has no permission for. Now the browser tries to read first few bytes of the file to check if it can actually be uploaded.
The second is showing a progress-bar when uploading the attachments.

### DIFF
--- a/app/assets/sass/components/_case-new.scss
+++ b/app/assets/sass/components/_case-new.scss
@@ -2,33 +2,35 @@
 	.config-options{
 		label.label-small-mar{
 			margin: 0 0 0;
-		}
-		label{
-			margin: 15px 0 0;
-		}
-		input{
-			margin: 5px 0 0;
-		}
-		div{
-			margin: 5px 0 0;
-		}
-		textarea{
-			margin: 5px 0 0;
-		}
+			}
+			label{
+				margin: 15px 0 5px;
+			}
+			input{
+				margin: 5px 0 0;
+			}
+			textarea{
+				margin: 5px 0 0;
+			}
+
+			#attach-file {
+				margin-top: 5px;
+				margin-bottom: 5px;
+			}
 	}
 	.rha-case-more-options{
 		label{
-			margin: 15px 0 0;
+			margin: 15px 0 5px;
 		}
 		input{
-			margin: 5px 0 0;
-		}
-		div{
 			margin: 5px 0 0;
 		}
 		.glyphicon-question-sign{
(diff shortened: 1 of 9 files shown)


app/assets/sass/components/_case-new.scss
```diff
@@ -2,34 +2,36 @@
...
-		}
-		label{
-			margin: 15px 0 0;
-		}
-		input{
-			margin: 5px 0 0;
-		}
-		div{
-			margin: 5px 0 0;
-		}
```
jtrantin: Removed this `div` rule, as it is too general. Replaced by variations of other rules.

jtrantin:
The build fails due to missing dependencies. Needs to be restarted once common and strata.js are merged and released.

engineersamuel:
Reviewed.  afaict the code looks good, get's a bit complicated with the angular digests and such, I'm assuming you've tested that well and it works.   Would be nice to see screens of the new error display and how things look.  I'll take one more glance at the code but nothing obviously stuck out as wrong.

jtrantin:
@engineersamuel  Unfortunately, the manual digests are needed, because lot of the attachment code is triggered from outside Angular. As for the error message, I don't have the screen right now, but it is regular danger message at the top saying `{{errorFileName}} cannot be read by the browser. Check your privileges and make sure you are allowed to read the file.`

However, I do have screens for the progress bars - https://goo.gl/photos/mtaw11DQmmCoEned9.

I have thoroughly tested this in Chrome, FF and IE11, so I hope all should work just fine.

app/cases/services/attachmentsService.js
```diff
@@ -51,54 +53,9 @@ export default class AttachmentsService {
                 this.originalAttachments = attachments;
             }
         };
-        this.postBackEndAttachments = function (caseId) {
```
gandhikeyur: Test cases are failing as it is not able to find this method. 
This method is getting called in the current file itself, in the updateAttachments function.
Please rectify this.

jtrantin: Oh! Not sure how I could have missed this. Thanks!

vrathee:
Reviewed!

PR #602, opened 2016-11-24
PCM-3721: Bumping common version

jtrantin:


jtrantin:
Will merge this without review, since it's just bumping a version of common in package.json.

PR #605, opened 2016-12-06
PCM-3784 code changes

gandhikeyur:
@jtrantin @renujhamtani Please review.

https://projects.engineering.redhat.com/browse/PCM-3784 

Thing to note:
I have added ".btn-primary" to those buttons that were always enabled.

However, if we added ".btn-primary" to buttons that get disabled conditionally, they appeared to be of light blue color, which is not desirable. Hence, I applied ".btn-primary" condition using ng-class only when the buttons are enabled.

### DIFF
--- a/app/cases/views/addCommentSection.jade
+++ b/app/cases/views/addCommentSection.jade
@@ -37,6 +37,7 @@ div(ng-show='EDIT_CASE_CONFIG.showAttachments && securityService.loginStatus.aut
       button#rha-case-addcommentbutton.btn.btn-app(
         ng-hide='addingComment || addingattachment',
         ng-disabled='!ie8 && !ie9 && (CaseService.disableAddComment && AttachmentsService.updatedAttachments.length === 0 && !AttachmentsService.hasBackEndSelections())',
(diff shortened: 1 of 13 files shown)


app/cases/views/addCommentSection.jade
```diff
@@ -37,6 +37,7 @@ div(ng-show='EDIT_CASE_CONFIG.showAttachments && securityService.loginStatus.aut
       button#rha-case-addcommentbutton.btn.btn-app(
         ng-hide='addingComment || addingattachment',
         ng-disabled='!ie8 && !ie9 && (CaseService.disableAddComment && AttachmentsService.updatedAttachments.length === 0 && !AttachmentsService.hasBackEndSelections())',
+        ng-class="{'btn-primary': !(!ie8 && !ie9 && (CaseService.disableAddComment && AttachmentsService.updatedAttachments.length === 0 && !AttachmentsService.hasBackEndSelections()))}",
```
jtrantin: We don't use the ie8 and ie9 flags, so I think you can remove them.

gandhikeyur:
Removed ie8 & ie9 occurences

PR #606, opened 2016-12-06
PCM-3752: Removed duplicate file upload in New view

jtrantin:
@gandhikeyur @renujhamtani Please review.

Note there is some weirdness in how strata is handling attachments in new cases (which also causes PCM-3769), so if you test something, make sure to test it on an older case. I will create strata jiras.

https://projects.engineering.redhat.com/browse/PCM-3752

### DIFF
--- a/app/cases/controllers/discussionSection.js
+++ b/app/cases/controllers/discussionSection.js
@@ -75,10 +75,6 @@ export default class DiscussionSection {
             CaseService.disableAddComment = true;
         };
 
-        if (securityService.loginStatus.isLoggedIn) {
-            DiscussionService.getDiscussionElements($stateParams.id);
-        }
-
         $scope.requestManagementEscalation = function () {
             $uibModal.open({
                 template: require('../views/requestManagementEscalationModal.jade'),
--- a/app/cases/views/new.jade
+++ b/app/cases/views/new.jade
@@ -88,39 +88,6 @@ div(rha-header='', page='newCase')
                                 div(ng-bind-html='$parent.AttachmentsService.parseArtifactHtml()')
                                 div#attachments(rha-listnewattachments='')
                                 div#attach-file(rha-attachlocalfile='', disabled='$parent.CaseService.submittingCase')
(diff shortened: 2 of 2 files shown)


app/cases/controllers/discussionSection.js
```diff
@@ -75,10 +75,6 @@ export default class DiscussionSection {
             CaseService.disableAddComment = true;
         };
 
-        if (securityService.loginStatus.isLoggedIn) {
```
jtrantin: This is already in `$scope.init`, so we were actually fetching the discussion elements twice.

app/cases/views/new.jade
```diff
@@ -88,39 +88,6 @@ div(rha-header='', page='newCase')
                                 div(ng-bind-html='$parent.AttachmentsService.parseArtifactHtml()')
                                 div#attachments(rha-listnewattachments='')
                                 div#attach-file(rha-attachlocalfile='', disabled='$parent.CaseService.submittingCase')
-                                .rha-bottom-border(ng-hide='$parent.NEW_CASE_CONFIG.isPCM')
-                                    .row.rha-create-field(ng-hide='$parent.CaseService.submittingCase')
-                                        .col-xs-12
-                                            div(ng-show='$parent.NEW_CASE_CONFIG.showServerSideAttachments')
```
gandhikeyur: @jtrantin 
Could you please discuss the use of NEW_CASE_CONFIG parameters with Spenser.
I am not sure where these flags are used, were they used for GS4 or for PCM to working as "InProduct"

jtrantin:
@gandhikeyur Talked with spenser yesterday, and both GS4 and In-Product shouldn't be a problem. GS4 is running off an old PCM version and we don't expect it to be retaged to a newer versions. This is similar for the In-Product, which is additionally deprecated.
So, we don't have to worry about these flags, actually, we can remove them altogether from the code along with the non-PCM specific code (which should simplify it a lot), I will create a JIRA for that.

PR #607, opened 2016-12-12
PCM-3818: Fixing Advanced Search init-related regressions

jtrantin:

https://projects.engineering.redhat.com/browse/PCM-3818

This was mostly caused by the session.js-related changes in securityService. Fix for that is already common module, while testing I found some other problems, so I fixed those too.

### DIFF
--- a/app/cases/controllers/advancedSearchController.js
+++ b/app/cases/controllers/advancedSearchController.js
@@ -52,10 +52,18 @@ export default class AdvancedSearchController {
             });
 
             if ($scope.solrQuery != null && SOLRGrammarService.parseSafe($scope.solrQuery) != null) {
-                $timeout(() => {
+                const search = () => {
                     AdvancedCaseSearchService.clearColumns(); // reset columns just before search is started to prevent UI glitches
                     $scope.submitSearch(); // Delay the search, so that rha-filterselect is initiated
-                }, 1000);
+                };
+                if(CaseService.filterSelect) {
+                    search();
+                } else {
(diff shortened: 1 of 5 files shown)


app/cases/controllers/advancedSearchController.js
```diff
@@ -52,10 +52,18 @@ export default class AdvancedSearchController {
             });
 
             if ($scope.solrQuery != null && SOLRGrammarService.parseSafe($scope.solrQuery) != null) {
-                $timeout(() => {
+                const search = () => {
                     AdvancedCaseSearchService.clearColumns(); // reset columns just before search is started to prevent UI glitches
                     $scope.submitSearch(); // Delay the search, so that rha-filterselect is initiated
-                }, 1000);
```
jtrantin: Now, we don't wait for some period of time until rha-filterselect does it's job, but instead filterselect fires an event when it initializes. This enables to chain these two initialization procedures in a much predictable way.

app/main.module.js
```diff
@@ -77,6 +77,12 @@ const app = angular.module('RedhatAccess', [
     'RedhatAccess.escalation'
 ]);
 
+app.run(["$rootScope",function ($rootScope) {
+    Promise.setScheduler(function (cb) {
+        $rootScope.$evalAsync(cb);
+    });
+}]);
```
jtrantin: Note sure where this code disappeared (I believe we had it previously), but we need it when using bluebird with Angular.

vrathee:
Reviewed!

jtrantin:
Bumped common version.

PR #612, opened 2017-01-10
PCM-3837: Checking if watcher is internal before adding him

jtrantin:
@vrathee @gandhikeyur @engineersamuel Please review.

As per the latest comment in https://projects.engineering.redhat.com/browse/PCM-3837, we are fetching the user info to get the `is_internal` flag. This is a short term fix, ultimately we'll want to add that flag into the account/users response.


engineersamuel:
Reviewed.

PR #615, opened 2017-02-02
PCM-4142: Fixing empty entitlements issue

jtrantin:
@engineersamuel Please review.

https://projects.engineering.redhat.com/browse/PCM-4142

The second part of the fix for PCM-4142. Theoretically, we don't need to add the empty-check, since it works fine with only the common changes. But I feel much better having two separate fixes in place :)


jtrantin:
Do not merge this yet. I want to find a reproducer in QA so that we can test this actually solves the problem.

PR #627, opened 2017-03-30
PCM-4383 Fix incorrect tag attribute ref

kunyan:


kunyan:
@jtrantin pls review

jtrantin:
@kunyan Check Swapna's comment in the JIRA. She found one more similar problem, so you can fix that too in this PR.

kunyan:
@jtrantin  fixed

PR #632, opened 2017-04-06
PCM-4392 Verbiage change on Create form to help clarify notification sending

kunyan:


kunyan:
@jtrantin @renujhamtani pls review

app/cases/views/new.jade
```diff
@@ -138,7 +138,7 @@ div(rha-header='', page='newCase')
                             div(ng-show="$parent.CaseService.showFts()")
                                 label#rha-24X7_Contact(translate,translate-comment="Noun") 24x7 Contact
                                 input#rha-24X7_Contact_input(ng-model="$parent.CaseService.fts_contact")
-                            label(for='email-notifications',translate) Send Email Notifications to
+                            label(for='email-notifications',translate) By opening this case you will automatically receive notifications. Send Additional Email Notifications to:
```
jtrantin: This text seems a bit too long for a label. Can you provide a screen how it looks like?

kunyan: https://projects.engineering.redhat.com/secure/attachment/36504/result.png

jtrantin:
Looks good, though I am slightly concerned how that long text will look like in a label. Maybe we could keep just "Send Additional Email Notifications to" and add an icon with a tooltip containing the rest.

kunyan:
result
I'm had modified it to this style, but according the request, they ask using full word as label

jtrantin:
I like the second version very very much! I think it's much clearer than putting all into label. Honestly, I think users are much more likely to read the text if it's formatted nicely, having a long bold text doesn't help readability.. @renujhamtani  What do you think? Can we go ahead with the version in the second screen, even though James Bailey and Jeremy Perry asked to put all the text into label? I think they didn't mean putting it into the label is a hard requirement, as long as the text is visible in the same general area.

renujhamtani:
I also liked the second screen.

@kunyan  Please put the screen in the JIRA and ask Jeremy Perry and Jamesh Bailey feedback on this. 
Also we need to implement this on both screens(case view page and case new page). So put both the screen in the JIRA.

renujhamtani:
@kunyan  Can you resolve the merge conflicts?

renujhamtani:
@kunyan we will wait for Jeremy or James reply.

kunyan:
I updated to using tooltip style. also fix following issues  in latest commit

- Fix Hostname icon no space issue
- Fix Send Additional Email user is loading, but chosen visible issue

PR #635, opened 2017-04-07
PCM-4396 Selected managed account name must be displayed in dropdown …

kunyan:
…once navigate back to case list page from other pages


kunyan:
@jtrantin pls review

PR #637, opened 2017-04-10
PCM-4417: Fixing account case filter and sort order

jtrantin:


jtrantin:
Also fixed sort order which was broken.

app/cases/services/searchCaseService.js
```diff
@@ -53,8 +53,6 @@ export default class SearchCaseService {
             caseGroup: null,
             accountNumber: null,
             searchString: null,
-            sortField: null,
-            sortOrder: null,
```
jtrantin: Removed sort-fields from searchParams. Sort order is already stored in local storage separately.

app/cases/services/searchCaseService.js
```diff
@@ -65,7 +63,25 @@ export default class SearchCaseService {
...
+            if (!RHAUtils.isEmpty(CaseService.filterSelect.sortOrder)) {
+                //This is a hack because strata returns the severities in reverse order
+                if ((CaseService.filterSelect.sortField === 'severity' || CaseService.filterSelect.sortField === 'status') && CaseService.filterSelect.sortOrder === 'ASC') {
+                    caseFilter.sort_order = 'DESC';
+                } else if ((CaseService.filterSelect.sortField === 'severity' || CaseService.filterSelect.sortField === 'status') && CaseService.filterSelect.sortOrder === 'DESC') {
+                    caseFilter.sort_order = 'ASC';
+                } else {
+                    caseFilter.sort_order = CaseService.filterSelect.sortOrder;
+                }
+            }
```
jtrantin: Restored the correct previous logic for sort-fields

PR #647, opened 2017-04-26
PCM-4466: Let external org-admins create case on their managed accounts

jtrantin:
@gandhikeyur Please review.

Surprisingly, the existing components ca be reused to full extent, I only changed when they are shown. Tested with internal/external org-admin/non-org-admin and all seems to be working fine.

### DIFF
--- a/app/cases/controllers/new.js
+++ b/app/cases/controllers/new.js
@@ -109,6 +109,9 @@ export default class New {
             CaseService.clearCase();
             RecommendationsService.clear();
             ProductsService.clear();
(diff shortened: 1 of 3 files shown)


vrathee:
On accountSelect.jade  we have added internal check :  This entire component is displayed inside a component that already has this check.. we dont let external users select account number.. so this check is redundant

vrathee:
@jtrantin Never mind.. I see that we have removed the external check.. Sorry for bugging... Reviewed!

PR #651, opened 2017-05-03
Pcm 4503 and PCM-4507 Fixes related to account number select and enhanced SLA on create case page

vrathee:
@renujhamtani @gandhikeyur Please review

### DIFF
--- a/app/cases/controllers/new.js
+++ b/app/cases/controllers/new.js
@@ -109,9 +109,6 @@ export default class New {
             CaseService.clearCase();
             RecommendationsService.clear();
             ProductsService.clear();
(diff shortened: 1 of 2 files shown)


app/cases/controllers/new.js
```diff
@@ -109,9 +113,6 @@ export default class New {
             CaseService.clearCase();
             RecommendationsService.clear();
             ProductsService.clear();
-            if(RHAUtils.isEmpty(CaseService.account.number)) {
-                CaseService.account.number = securityService.loginStatus.authedUser.account_number;
-            }
```
vrathee: The only change for https://projects.engineering.redhat.com/browse/PCM-4503

PR #680, opened 2017-08-04
PCM 5052 : reading object from url and filling new case fields

gandhikeyur:
@engineersamuel @vrathee @kunyan @renujhamtani  Please review.

https://projects.engineering.redhat.com/browse/PCM-5052 

As of now, I have hardcoded a few lines to manually put a new case object in the localStorage cache for testing.

**Note :** I would be removing it in once the solution engine or container catalog are ready with their changes.

The flow would be as furnished below:

1) Solution engine & Container Catalog to store case creation details in local storage and pass the local storage key as the paramter in the new case url like https://access.qa.redhat.com/support/cases/#/case/new?caseCreateKey=cc-123456

2) PCM to read this caseCreateKey and fetch the stringified object from localStorage cache, parse it to get actual JSON object and fill the details in the respective fields on the case create page.

3) For Container Catalog, PCM would be removing the localStorage cache object right after populating the case fields. 
However, for Solution engine, PCM would have to retain the localStorage cache object till the user creates the case and the case number has to be passed to their API along with some more details(specificallu GUID) that they would have stored in the localStorage cache object. Once this API request is submitted and response is received, PCM would be deleting the localStorage cache object that is received via Solution engine.

More details on the approach can be found @ https://docs.google.com/document/d/1uA773GmMayhpSW0QfckPgiqlPvlAwS0dPbslCf-pJZc/edit

### DIFF
--- a/app/cases/controllers/new.js
+++ b/app/cases/controllers/new.js
@@ -220,6 +220,24 @@ export default class New {
             }
         };
 
+        // This function has been added for the Solution Engine & Container Catalog
+        // Solution Engine or Container Catalog will send the key in URL parameter and store the new case details object in localStorage
+        // PCM reads the localStorageCache using the caseCreateKey passed in the URL params and populates case fields accordingly
+        // PCM-5052
+        $scope.getCaseDetailsFromLocalStorage = () => {
+            if (RHAUtils.isNotEmpty(CaseService.localStorageCache) && CaseService.localStorageCache.get(CaseService.externalCaseCreateKey)) {
+                var draftNewCase = JSON.parse(CaseService.localStorageCache.get(CaseService.externalCaseCreateKey));
+                CaseService.kase.problem = draftNewCase.problemStatement;
+                CaseService.kase.environment = draftNewCase.environment;
+                CaseService.kase.occurance = draftNewCase.frequency;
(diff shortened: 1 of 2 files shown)


app/cases/controllers/new.js
```diff
@@ -307,6 +325,21 @@ export default class New {
             if (RHAUtils.isNotEmpty(urlParameter.product)) {
                 $scope.setProductAndVersion(urlParameter.product, urlParameter.version);
             }
+            if (RHAUtils.isNotEmpty(urlParameter.caseCreateKey)) {
+                CaseService.externalCaseCreateKey = urlParameter.caseCreateKey;
+                //dummy stringified case object, need to delete this once get original object from solution engine or container catalog
+                //delete the below 2 lines when original integration is complete.
+                let newStrigifiedCaseObject = '{"product":"Red Hat Enterprise Linux","version":"6.0","problemStatement":"test problem","issue":"test issue","environment":"test env","frequency":"test frequency","businessImpact":"test business","guid":"test"}';
+                CaseService.localStorageCache.put(CaseService.externalCaseCreateKey, newStrigifiedCaseObject);
+
```
gandhikeyur: As mentioned in the PR description, above lines are hard coded as of now and would be removed when the integration is live.

PR #687, opened 2017-08-16
PCM 5159 hiding solution engine message on case list page

gandhikeyur:
@renujhamtani @engineersamuel @kunyan Please review.

As per the comment by Tom Butt in PCM-5159, hiding this message until we receive an ack from Solution Engine stakeholders.

https://projects.engineering.redhat.com/browse/PCM-5159


gandhikeyur:
Merging this as this is a small change. Please comment if you feel otherwise.

PR #719, opened 2018-01-16
Add OpenShift v3 config

kunyan:
* Will have a new Dockerfile based on IT bless images
* Will using a new version of expess as server application
* It won't impact anything for legacy Openshift v2


kunyan:
@gandhikeyur the current QA environment is using this configuration

PR #722, opened 2018-01-24
PCM-5590 && PCM-5589 ice/partner escalation form to use new object .

anujsi:


app/escalation/controllers/escalationRequest.js
```diff
@@ -12,6 +12,7 @@ export default class EscalationRequest {
         $scope.ESCALATION_TYPE = ESCALATION_TYPE;
         $scope.partnerGeoList = ['NA', 'EMEA', 'LATAM', 'APAC'];
         $scope.iceGeoList = ['NA', 'EMEA', 'LATAM', 'APAC', 'Combo'];
+        $scope.severityList = [1, 2, 3, 4];
```
engineersamuel: This may be expected in the ICE, I'm unsure, but are we mapping the severity as an integer vs. a string like '1 (Low)' ?

engineersamuel: I think I see now, the payload expects an integer value when creating an ice. Makes sense.  Just checking though, as really it makes little difference in showing 1-4 vs. a string.

anujsi:
@engineersamuel @gandhikeyur i am merging this PR , if you have any suggestion i will merge in another PR

PR #725, opened 2018-01-31
PCM-3841 Make selecting a case group required at the time of case open

aksgupta14:
If "require_cgroup_on_create" is returning true then user has to select a case group, othewise the submit button will remain disabled in create new case.

@gandhikeyur @anujsi @kunyan please review

### DIFF
--- a/app/cases/controllers/new.js
+++ b/app/cases/controllers/new.js
@@ -37,7 +37,7 @@ export default class New {
         $scope.recommendationsPerPage = 6;
         $scope.noEnhancedSLAMessage = gettextCatalog.getString("There are no remaining enhanced SLA's available");
 
-        $scope.$watch('CaseService.account.name', async function () {
+        $scope.$watch('CaseService.account.name', function () {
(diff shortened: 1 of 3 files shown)


app/cases/controllers/new.js
```diff
@@ -26,15 +26,27 @@ export default class New {
...
         $scope.kase = {
             notifiedUsers : []
         };
+        $scope.caseGroupRequired = false;
 
         $scope.isControlGroup = true;
         $scope.recommendationsPerPage = 6;
         $scope.noEnhancedSLAMessage = gettextCatalog.getString("There are no remaining enhanced SLA's available");
+
+        $scope.$watch('CaseService.account.name', async function () {
```
engineersamuel: I don't think this function would actually need to be marked async unless there is a corresponding await in the function body.

aksgupta14: Yep! Removed in the latest push. Thanks!

anujsi:
reviewed.

PR #728, opened 2018-03-06
PCM-6452 completed the jira with acceptance crietria

aksgupta14:
Applied logic when user require case group is option and when it isn't. Especially for managed account

@gandhikeyur @anujsi @engineersamuel please review

### DIFF
--- a/app/cases/controllers/new.js
+++ b/app/cases/controllers/new.js
@@ -37,7 +37,21 @@ export default class New {
         $scope.recommendationsPerPage = 6;
         $scope.noEnhancedSLAMessage = gettextCatalog.getString("There are no remaining enhanced SLA's available");
 
+        let isManagedAccount = false;
+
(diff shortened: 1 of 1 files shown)


app/cases/controllers/new.js
```diff
@@ -47,6 +61,38 @@ export default class New {
...
+                    $scope.caseGroupRequired = false;
+                    CaseService.requireCaseGroup = false;
+                }
+            }
+        });
+
+        $scope.$watch('CaseService.owner', function(){
+            if(isManagedAccount){
+                for(var a of CaseService.loggedInAccountUsers){
+                    if(CaseService.owner === a.sso_username){
```
engineersamuel: This assumes CaseService.owner is a string and the sso ? Just making sure you checked it is.

aksgupta14: Yes it's a string.

PR #736, opened 2018-03-16
PCM-5697 - Distinguish between partner and customer comments in partn…

anujsi:
@engineersamuel @gandhikeyur @aksgupta14 @kunyan Please review.
https://projects.engineering.redhat.com/browse/PCM-5697

### DIFF
--- a/app/cases/views/discussionSection.jade
+++ b/app/cases/views/discussionSection.jade
@@ -31,7 +31,7 @@ ul#tab_list.nav.nav-tabs(role="tablist")
                 .discussion-element(id="{{element.id}}", ng-repeat="element in DiscussionService.discussionElements | orderBy:'sortModifiedDate':commentSortOrder", ng-if='!element.draft')
                     article.comment.article-comment(ng-if="!element.file_name && element.comment_type !== 'chat' && !element.originating_system && (element.public === undefined || element.public === true)")
                         i.fa.fa-comments-o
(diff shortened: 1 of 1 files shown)


app/cases/views/discussionSection.jade
```diff
@@ -31,7 +31,7 @@ ul#tab_list.nav.nav-tabs(role="tablist")
                 .discussion-element(id="{{element.id}}", ng-repeat="element in DiscussionService.discussionElements | orderBy:'sortModifiedDate':commentSortOrder", ng-if='!element.draft')
                     article.comment.article-comment(ng-if="!element.file_name && element.comment_type !== 'chat' && !element.originating_system && (element.public === undefined || element.public === true)")
                         i.fa.fa-comments-o
-                        h4.meta {{'Message'|translate}}
+                        h4.meta {{'Message'|translate}}&nbsp; {{'('}}{{element.created_by_type}}{{')'}}
```
engineersamuel: I may be wrong, but wouldn't the following be simpler?

`h4.meta {{'Message'|translate}}&nbsp; ({{element.created_by_type}})`

I don't think angular will interpolate the ( and ) if it is just a character not enclosed in {{}}

anujsi: Yes, you are right angular interpolate the ( and ). I have made changes as per your suggestion.

PR #743, opened 2018-04-17
PCM-5730 - Partner query should search client account .

anujsi:
@kunyan @gandhikeyur @engineersamuel @aksgupta14 Please review.
https://projects.engineering.redhat.com/browse/PCM-5730

### DIFF
--- a/app/cases/services/searchCaseService.js
+++ b/app/cases/services/searchCaseService.js
@@ -171,6 +171,7 @@ export default class SearchCaseService {
         this.doCaseSearch = function() {
             const deferred = $q.defer();
             let sortField = CaseService.filterSelect.sortField === 'owner' ? 'contactName' : CaseService.filterSelect.sortField;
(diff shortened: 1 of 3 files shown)


anujsi:
@kunyan I have made changes as per your suggestion. Please review.

app/cases/services/searchCaseService.js
```diff
@@ -171,6 +171,7 @@ export default class SearchCaseService {
         this.doCaseSearch = function() {
             const deferred = $q.defer();
             let sortField = CaseService.filterSelect.sortField === 'owner' ? 'contactName' : CaseService.filterSelect.sortField;
+            const partnerSearch = RHAUtils.isEmpty(this.searchParameters.accountNumber) && this.isManagedAccount() ? true : false;
```
kunyan: I mean there is no need to using conditional operator in this case, because `isEmpty()` and `isManagedAccount()` will return boolean value, right ? we can just using it

anujsi:
@kunyan thanks for sharing feedback, I have made changes as per your suggestion.

PR #755, opened 2018-07-03
pcm-6621 - removing auto select for unsupported products

gandhikeyur:
@anujsi @aksgupta14 @kunyan Please review.

Removing preferred service level as we do not want the UI to auto select an entitlement if the product is not supported.

https://projects.engineering.redhat.com/browse/PCM-6621


anujsi:
Reviewed.

PR #767, opened 2018-08-02
PCM-7186 add spinner while performing RME request .

anujsi:
@vrathee @aksgupta14 Please review.
https://projects.engineering.redhat.com/browse/PCM-7186


vrathee:
@anujsi How about adding (ng-show='!submittingRequest') on the row itself and add another row(ng-show='submittingRequest') with a spinner in a column.. This way you will not be adding the same '!submittingRequest' to all the columns

anujsi:
@vrathee I have made changes as per your suggestion Please check.

PR #778, opened 2018-09-10
PCM-7326: Added conditional use of s3 upload based on account number

jeudy100:


app/cases/services/attachmentsService.js
```diff
@@ -155,9 +155,10 @@ export default class AttachmentsService {
 
         this.useS3Upload = async function() {
             const s3Configurations = await hydrajs.maintenance.getMaintenanceMode(this.s3Constants.s3_configurations);
-            const listObj = _.find(s3Configurations, { fieldName: this.s3Constants.accountWhitelist });
-            const whitelist = listObj.fieldValue.split(',');
             const uploadConfig = _.find(s3Configurations, { fieldName: this.s3Constants.s3UploadFunctionality });
+            const lists = _.filter(s3Configurations, (val) => val.fieldName.indexOf(this.s3Constants.accountWhitelist) > -1);
+            const whitelist = [];
+            lists.forEach((val) => whitelist.push(...val.fieldValue.split(',')));
```
vrathee: You can also use reduce here.

app/cases/services/attachmentsService.js
```diff
@@ -155,9 +155,10 @@ export default class AttachmentsService {
 
         this.useS3Upload = async function() {
             const s3Configurations = await hydrajs.maintenance.getMaintenanceMode(this.s3Constants.s3_configurations);
-            const listObj = _.find(s3Configurations, { fieldName: this.s3Constants.accountWhitelist });
-            const whitelist = listObj.fieldValue.split(',');
             const uploadConfig = _.find(s3Configurations, { fieldName: this.s3Constants.s3UploadFunctionality });
+            const lists = _.filter(s3Configurations, (val) => val.fieldName.indexOf(this.s3Constants.accountWhitelist) > -1);
```
vrathee: indexOf will mean whiteList can be anywhere in the key.. but as oer your JIRA comment you want it only in the end as a surfix. Thus, might need to change here to make sure something unwanted is not included.

app/cases/views/attachLocalFile.jade
```diff
@@ -1,7 +1,7 @@
 .row
   .col-xs-12
     button.btn.btn-primary.btn-app(ng-click='getFile()', ng-disabled='disabled', translate='', ng-hide='AttachmentsService.updatedAttachments.length == 0') Attach Another File
-    button.btn.btn-primary.btn-app(uib-tooltip="{{'Can now accept large attachments (~5TB)' | translate}}" ng-click='getFile()', ng-disabled='disabled', translate='', ng-hide='AttachmentsService.updatedAttachments.length > 0') Attach File
+    button.btn.btn-primary.btn-app(uib-tooltip="{{attachFileTT | translate}}" ng-click='getFile()', ng-disabled='disabled', translate='', ng-hide='AttachmentsService.updatedAttachments.length > 0') Attach File
```
vrathee: Was this message something that was requested in PCM JIRA?

jeudy100: yes it was

app/cases/services/attachmentsService.js
```diff
@@ -146,7 +148,97 @@ export default class AttachmentsService {
             this.loading = false;
         };
 
-        this.updateAttachments = async function (caseId) {
+        this.updateAttachments = async function(caseId) {
+            try {
+                if (!this.s3AccountConfigurations) {
+                    this.s3AccountConfigurations = await hydrajs.kase.attachments.getS3UploadAccounts();
```
vrathee: Here to this will happen for each case load.. The s3 configs will not change for each case and can be loaded once when app loads.. see how we get pcm_maintainance configs.. this might nee changes to https://github.com/redhataccess/redhat_access_pcm_ascension_common/

vrathee:
Reviewed!

PR #787, opened 2018-09-19
PCM-7369: fixed attachments not downloading correctly

jeudy100:


jeudy100:
@vrathee it turns out that strata attachments have both link and uri while s3 attachments only have link, so I changed it to just use element.link

jeudy100:
@vrathee MR https://github.com/redhataccess/redhat_access_angular_ui/pull/786/files needs to be merged before this works b/c i added the new hydrajs in that commit

PR #800, opened 2018-09-25
Fixing 401 error call for s3config, executing call after init success

gandhikeyur:
@vrathee @anujsi @jeudy100 Please review

While debugging with najain, we observed that the s3 config calls were getting executed even before the init of the app and hence the cookies were not getting parsed in the Api call which resulted in 401 errors randomly for s3 configs.
Having s3 config result is much needed for the new uploader functionality to work.

I did not have a jira number handy, hence I have not tagged any jira in this PR.


anujsi:
@gandhikeyur  init() method will call 2 times. did you test it locally?

anujsi:
Reviewed.

PR #809, opened 2018-10-03
Not sending Authorization header in IE11

aksgupta14:
@anujsi @gandhikeyur @vrathee please review


anujsi:
Reviewed.

vrathee:
@akshat have you tried your calls to systems other then hydra?

Also, we should remove the credentials omit header for ie11.. so be sure we don't break anything unintentionally..

vrathee:
As discussed, it will work as expected.. Reviewed!

PR #817, opened 2018-10-15
PCM-6913 Added case group

gakshat14:
https://projects.engineering.redhat.com/browse/PCM-6913
@anujsi @gandhikeyur @vrathee @asonawan please review.

### DIFF
--- a/app/cases/views/list.jade
+++ b/app/cases/views/list.jade
@@ -48,13 +48,15 @@ div(ng-hide='showCaseList() || securityService.loginFailure || !securityService.
                             .case-id
                                 .subheading {{case.case_number}}
                                 .description {{case.status}}
-                    .col-sm-8.col-md-6.item-group
+                    .col-sm-8.col-md-4.item-group
                         .subheading
                             | {{case.summary}}
                         .description
                             | {{case.product}}
                             span
                             | &nbsp;{{case.version}}
+                    .col-md-2.item-group.change-log
+                            span(ng-show='securityService.loginStatus.authedUser.org_admin') {{case.folder_name}}
                     .col-sm-12.col-md-2(ng-show='CaseService.filterSelect.sortField === "lastModifiedDate"')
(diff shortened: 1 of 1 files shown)


anujsi:
Reviewed.

app/cases/views/list.jade
```diff
@@ -48,13 +48,15 @@ div(ng-hide='showCaseList() || securityService.loginFailure || !securityService.
...
-                    .col-sm-8.col-md-6.item-group
+                    .col-sm-8.col-md-4.item-group
                         .subheading
                             | {{case.summary}}
                         .description
                             | {{case.product}}
                             span
                             | &nbsp;{{case.version}}
+                    .col-md-2.item-group.change-log
+                            span(ng-show='securityService.loginStatus.authedUser.org_admin') {{case.folder_name}}
```
vrathee: @aksgupta14 
1. Shouldn't we show/hide the entire col instead of the content based on weather user is org_admin or not?  ie. the col layout should only change for org_admins. Also, why didn't we modify the small screen resolution layout..  (col-sm-)
2. Isn't there a column header that we need to add?

gakshat14: 1. showing/hiding the entire col was making the whole UI look weird. I didn't modify because while testing it was working fine in responsive mode as well.

2. It's not a table, and adding column header will require to redesign this whole thing. Also had conversation with Akshay and he was fine with this change.

akkias: > @aksgupta14
> 
> 1. Shouldn't we show/hide the entire col instead of the content based on weather user is org_admin or not?  ie. the col layout should only change for org_admins. Also, why didn't we modify the small screen resolution layout..  (col-sm-)
Since it is not Table, it will change the alignment of the rows based on content so, whatever Akshat has done is good IMO.
> 2. Isn't there a column header that we need to add?
We have thought of this option but then we will have to change many things here to make it work.

PR #818, opened 2018-10-15
PCM-7521: fixed post button appearing during download

jeudy100:


vrathee:
@jeudy100 Reviewed and merging this one...

However I have a suggestion.. we should remove the aborted file from the attachment list so that even if the button is visible its not uploaded again...

PR #822, opened 2018-10-25
PCM-7544: fixed bug when removing attachment with index after an abort

jeudy100:


app/cases/services/attachmentsService.js
```diff
@@ -153,14 +154,10 @@ export default class AttachmentsService {
...
-                const response = await this.isValidS3UploadAccount() ? await this.updateAttachmentsS3(caseId) : await this.updateAttachmentsStrata(caseId);
-                this.uploadingAttachments = false;
-                return response;
-            } catch (error) {
-                this.uploadingAttachments = false;
-            }
+            this.uploadingAttachments = true;
+            const response = await this.isValidS3UploadAccount() ? await this.updateAttachmentsS3(caseId) : await this.updateAttachmentsStrata(caseId);
+            this.uploadingAttachments = false;
+            return response;
```
vrathee: Why did we remove it from a try catch block?

vrathee:
The changes look good from coding perspective but cannot make sure this is all we need to get this working... Will need to be tested by QE team for that.

Reviewed!

PR #834, opened 2018-11-14
PCM-7319 - Internal users can view case owners contact information

anujsi:
https://projects.engineering.redhat.com/browse/PCM-7319

### DIFF
--- a/app/cases/views/edit.jade
+++ b/app/cases/views/edit.jade
@@ -23,7 +23,7 @@ div.edit(ng-if='showCasePage()')
       .row.row-very-short
         span {{'Assigned to'|translate}} {{$parent.CaseService.kase.owner}}
         span &nbsp;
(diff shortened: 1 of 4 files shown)


app/cases/views/edit.jade
```diff
@@ -22,6 +22,8 @@ div.edit(ng-if='showCasePage()')
         span(id="rha-case-filedOn" translate)  Filed on {{$parent.CaseService.kase.created_date}} by {{$parent.CaseService.kase.created_by}}
       .row.row-very-short
         span {{'Assigned to'|translate}} {{$parent.CaseService.kase.owner}}
+        span &nbsp;
+        span(ng-if='$parent.securityService.loginStatus.authedUser.is_internal && $parent.securityService.isCep').glyphicon.glyphicon-info-sign(uib-popover-html="ownerTooltip", popover-placement="right")
```
vrathee: Does this needs to be related to CEP? I think we can start showing owner to internal users even before that.

@gandhikeyur Thoughts?

anujsi: @vrathee As per discussed with Haroon we don't need CEP flag configurable for this feature.

PR #840, opened 2018-11-27
PCM-7629 - fix minor issue for case owner tooltip .

anujsi:
@vrathee @aksgupta14 Please review.
https://projects.engineering.redhat.com/browse/PCM-7629

### DIFF
--- a/app/cases/controllers/edit.js
+++ b/app/cases/controllers/edit.js
@@ -21,6 +21,7 @@ export default class Edit {
         $scope.isShowRmeEscalationBox = false;
         $scope.isCreateRmeEscalationBox = true;
         $scope.ownerTooltip = '';
+        $scope.isShowOwnerTooltip = true;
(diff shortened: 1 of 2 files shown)


app/cases/controllers/edit.js
```diff
@@ -21,6 +21,7 @@ export default class Edit {
         $scope.isShowRmeEscalationBox = false;
         $scope.isCreateRmeEscalationBox = true;
         $scope.ownerTooltip = '';
+        $scope.isShowOwnerTooltip = false;
```
vrathee: Rename to $scope.showOwnerTooltip

vrathee: In fact we don't need this variable, we can check if ownerTooltip is blank and show hide accordingly.

anujsi:
@vrathee I have made changes as per your suggestion. Please check.

vrathee:
Reviewed!

PR #852, opened 2018-12-18
Comments improvements

praniredhat:
@vrathee  @gandhikeyur Please review,

### DIFF
--- a/app/cases/controllers/discussionSection.js
+++ b/app/cases/controllers/discussionSection.js
@@ -39,18 +39,27 @@ export default class DiscussionSection {
             }
         ];
         $scope.DiscussionService = DiscussionService;
-        
+
         $scope.$on(CASE_EVENTS.searchSubmit, function () {
-            console.log("search:", SearchBoxService.searchTerm);
-            DiscussionService.doSearch(SearchBoxService.searchTerm);
+            DiscussionService.doSearch(SearchBoxService.searchTerm, $scope.attachments);
         });
 
+        $scope.$on(CASE_EVENTS.searchBoxChange, function () {
+            // when search box is cleared via clear icon
+            if (!SearchBoxService.searchTerm) {
+                DiscussionService.doSearch(SearchBoxService.searchTerm, $scope.attachments);
+            }
+        });
 
         $scope.getPaginationData = (pageSize, currentPage) => {
             $scope.pageSize = pageSize;
             $scope.currentPage = currentPage;
-            DiscussionService.highlightSearchResults(SearchBoxService.searchTerm);
+            $scope.onChange();
         };
+
+        $scope.onChange = () => {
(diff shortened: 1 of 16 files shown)


praniredhat:
foo

praniredhat:
@asonawan Please have a look into UI changes and screenshot attached.

app/cases/controllers/discussionSection.js
```diff
@@ -38,23 +36,74 @@ export default class DiscussionSection {
...
+        }
+
+        $scope.scrollToTopOfAttachmentsSection = () => {
+            let element = document.getElementById('top-attachments-section');
+            element && element.scrollIntoView(true);
+        }
+
+        $scope.scrollToBottomOfAttachmentsSection = () => {
+            let element = document.getElementById('bottom-attachments-section');
+            element && element.scrollIntoView(true);
```
vrathee: you can create a function that takes an id and use it directly by passing desired id and scrolling to that element..

app/cases/controllers/discussionSection.js
```diff
@@ -319,5 +386,25 @@ export default class DiscussionSection {
         $scope.$watch('CaseService.account.number', async () => {
             await AttachmentsService.reEvaluateS3EnabledForAccount();
         }, true);
+
+        $scope.mouseOver = (element) => {
+            const isUserSearching = !!SearchBoxService.searchTerm;
+            if (isUserSearching) $scope.showJumpToComment = element.id;
```
vrathee: change to $scope.showJumpToComment = !!element.id so that you have an actual boolean value

app/cases/controllers/discussionSection.js
praniredhat: done :) @vrathee

app/cases/controllers/discussionSection.js
praniredhat: I needed id to be captured, to know which comment should show label "jump to comment" on hover. I may change the variable name with much appropriate meaning or any other suggestion you have to acheive this ?

I used falsy in mouse out it just reflects that no comment Id is selected @vrathee

vrathee: Yes we should change the variable name and also set undefined/null for falacy just like normal non-boolean variables

praniredhat:
Fixing tests.

app/cases/controllers/discussionSection.js
praniredhat: okay cool.

anujsi:
Reviewed.

vrathee:
Apart from moving page selector before the pagination.. I don't have any other comment.. in fact I would want @asonawan opinion on where it should go..

You can take care of it in a separate PR if needed.. Merging this one

PR #856, opened 2019-01-08
on delete attachment it should refresh list and on new comment post i…

praniredhat:
…t should clear search highlight


vrathee:
@praniredhat I see changes in only one file.. Is that it?

praniredhat:
@vrathee yes, one file change

PR #857, opened 2019-01-14
[PCM-2020] Linkifies bugzilla ID and simplifies linking process

praniredhat:
[JIRA] https://projects.engineering.redhat.com/browse/PCM-2020  
Please review @ascension @vrathee @anujsi

### DIFF
--- a/app/cases/services/linkifyService.js
+++ b/app/cases/services/linkifyService.js
@@ -5,25 +5,17 @@ export default class LinkifyService {
         this.linkifyWithCaseIDs = function (text) {
             if (RHAUtils.isEmpty(text)) return '';
 
-            var caseIdRegex = /([\s\('"]|^)(\d{8})([\s\)'"\.]|$)/g;
-            var caseLinkRegex = /<a href="#\/case\/\d{8}">\d{8}<\/a>/g;
-            var caseLinksinkifiedText = text.replace(caseIdRegex, '$1<a href="#/case/$2">$2</a>$3');
-            var textSegments = caseLinksinkifiedText.split(caseLinkRegex);
-            var caseLinks = caseLinksinkifiedText.match(caseLinkRegex);
-            // merge the elements back together and linkify textSegments
-            try {
-                var result = [$filter('linky')(textSegments[0], '_blank')];
-                for (var i = 1; i < textSegments.length; i++) {
-                    result.push(caseLinks[i - 1]);
-                    result.push($filter('linky')(textSegments[i], '_blank'));
-                }
-                return result.join('');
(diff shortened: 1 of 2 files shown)


app/cases/services/linkifyService.js
```diff
@@ -5,25 +5,17 @@ export default class LinkifyService {
...
-            }
+            var caseIdRegex = /([\s\('";]|^)(\d{8})([\s\)'"\&]|$)/g;
+            let caseLinksinkifiedText = text.replace(caseIdRegex, '$1<a href="#/case/$2">$2</a>$3');
+            return caseLinksinkifiedText;
+        }
+
+        this.linkifyBZIDs = function (text) {
+            if (RHAUtils.isEmpty(text)) return '';
 
+            var bzIdRegex = /(Bug|BZ)([\s\('";#]|^)(\d{7,})([\s\)'"\&]|$)/ig;
```
vrathee: @praniredhat  \s matches any whitespace character (equal to [\r\n\t\f\v ]) but what we want to match is just a space. Also, we don't want to match & in the end. And even if there is any of `"'#(` after a space, we want to match that too. 
I think a better regex would be `/(Bug|BZ)([[:blank:]])?([[:blank:]\('"#-])(\d{7,})([\)'"])?/ig`
it will match 
```
BZ(1234567) BZ (1234567)
BZ"1234567"
BZ "1234567"
BZ'1234567'
BZ '1234567'
BZ#1234567
BZ #1234567
BZ-1234567
```

praniredhat: @vrathee   

Did not get, why we should not match all whitespace characters. 

The steps for parsing comment text are such that  
text (which has links in the format of http://foo.com)  => linky filter => html (with transformed links into anchor tags and rest of the text gets sanitized and performs html [encoding](https://en.wikipedia.org/wiki/Character_encodings_in_HTML) encoded characters starts with & and end with ;
That's why I had to include & and ;  

And thanks for the given list of supported formats, will make sure they work and better would be if I write unit test for them.

vrathee: @praniredhat We are only interested in space and tab between the words.. \s also validates if there is a newLine

You can validate your regex on https://regex101.com/  Its very informative and helpful utility.

praniredhat: Made changes

praniredhat:
@vrathee  Changed Regex shall I merge this?  Is there a way can i write test for service, not sure how to create an instance of LinkifyService service.

vrathee:
@praniredhat The regex doesn't cover the ending brackets and quotes..

PR #861, opened 2019-01-22
Fix for s3 upload

aksgupta14:
@vrathee @anujsi @gandhikeyur please review


vrathee:
@aksgupta14 Please remove your hydrajs changes.. they are not needed as part of hotfix.. infact some of them are not at all needed

PR #865, opened 2019-01-25
Hotfix 1.4.39 => master

vrathee:
@praniredhat @anujsi Please review


chopraapooja:
LGTM

chopraapooja:
@vrathee  Are we also displaying createdbyUser ? or rather should we ?

vrathee:
@praniredhat This is a temporary change and will be reverted next week.. thus we decided not to increase the scope of testing.. modifiedBy user was not impacted by the SFDC change and attachments are hardly modified.. So we should be good.

PR #877, opened 2019-02-13
PCM-7190: Fixes padding on case edit

praniredhat:
Please review @asonawan @vrathee


praniredhat:
Merging this @asonawan @vrathee

PR #878, opened 2019-02-15
While upload is in progress, Keep trying to refreshing Auth token after every 3 minutes if its ie11 and token is set to expire within 50 seconds.

vrathee:
@praniredhat Please review


vrathee:
@praniredhat Merging this in the interest of time..

praniredhat:
Reviewed.

PR #879, opened 2019-02-25
PCM-7730: Allow severity based on selected product and version.

praniredhat:
https://projects.engineering.redhat.com/browse/PCM-7730  
Please review: @anujsi @vrathee


app/cases/controllers/new.js
```diff
@@ -258,6 +258,7 @@ export default class New {
                 AlertService.addStrataErrorMessage(error);
             });
             $scope.groupsLoading = true;
+            ConfigService.loadConfig();
```
rh-kgarg: Do we need to load this config only when we create a new case?
I think we can load the config on app load.. If we start consuming configs in other pages too.WDYT?

rh-kgarg:
Rest looks good.

rh-kgarg:
Also you tagged some other Karan in the PR...:D

app/cases/controllers/new.js
praniredhat: Umm, I need a bit help here... if you see this new.js is a top component which handles /cases/new route, so what could be the best place to load config for entire cases.module should i write init function in cases.module.js?  @rh-kgarg

anujsi:
@praniredhat we are already loading `pcm_configurations` in the redhat_access_pcm_ascension_common project. try to use that configuration so you don't need to load it again.

praniredhat:
Hello,

Moved config service to commons, and now we have config loaded at one place during login init.

PR #902, opened 2019-04-23
PCM-7748: Added jump to attachment when searching in case discussion

jeudy100:
@vrathee @praniredhat  please review


vrathee:
Reviewed!

PR #904, opened 2019-04-23
PCM-8026: fixed severity selection after switching product

jeudy100:
@praniredhat @vrathee please review. For this, every time you switch a product it defaults to the lowest severity. I assumed that the lowest severity will always be an available choice, is this correct/acceptable functionality? if it isn't, I'll need to go back and implement a check for the enabled severities.

### DIFF
--- a/app/cases/controllers/severitySelect.js
+++ b/app/cases/controllers/severitySelect.js
@@ -5,10 +5,11 @@ export default class SeveritySelect {
         'ngInject';
 
         // INIT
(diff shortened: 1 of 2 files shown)


vrathee:
@jeudy100 We recently implemented a JIRA as part of which certain severities are disabled based on selected products. @praniredhat Can help you better as she worked on that JIRA

jeudy100:
@vrathee I added the check to get the default/valid severity. It will always default to the first severity in the list of valid severities.

PR #907, opened 2019-04-24
PCM-8058: added asterisk to severity icon when it has RMEs

jeudy100:
@anujsi @vrathee please review

### DIFF
--- a/app/assets/sass/components/_case-list.scss
+++ b/app/assets/sass/components/_case-list.scss
@@ -58,6 +58,15 @@ $subheading-margin-bottom: 4px;
 			background: #f0f0f0;
 		}
 	}
+
+    .escalated {
+        text-align: center;
+
+        & :first-child {
+            color: initial;
+        }
+    }
+
 }
 
 .bulk-closure.case-list {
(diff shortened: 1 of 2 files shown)


vrathee:
Reviewed!

praniredhat:
@jeudy100 Is the given solution `asterisk beside severity to represent this case is escalated` is approved by UX team ( @asonawan ) ?

Because to me this asterisk seems to have hidden meaning. How will user know what does it mean ?

jeudy100:
@praniredhat I have not and you are right. I could put a tooltip around the icon that will bring up a message, but that could be a little much if there are a lot of "asterisked" cases. I could add another column called "Escalated" and it could be just a checkmark to keep the column small (@asonawan)?

vrathee:
@jeudy100 I have commented on the JIRA to get clarity on how it should look. Holding off the merge till then.

jeudy100:
@vrathee @praniredhat I added the column and a tooltip. Inside the column you will either have a a check or an x. The tooltip is there because there are no column headers to explain what it actually is and just having a x or check could still be confusing as to what it defines. I know in the jira @vrathee you said to just leave it blank when there are no escalations, but I thought it would be confusing to  just have an empty column.

PR #910, opened 2019-05-02
PCM-5543: fixed xml not showing up by not binding it to the html

jeudy100:
@vrathee @praniredhat please review


vrathee:
Reviewed!

PR #918, opened 2019-05-06
Fix bug for FF where only host is referrer & better msg for Attachment description

praniredhat:
Please review @anujsi  @vrathee


anujsi:
Reviewed.

PR #920, opened 2019-05-07
PCM-7979 and PCM-7984 Fix - Provide Ability to remove added RH Watcher on Case Create Page

vrathee:
@anujsi @praniredhat Please Review


anujsi:
Reviewed.

PR #937, opened 2019-06-18
PCM-3329: added filters mentioned in PCM-3329, 8045 and 7855

jeudy100:
@anujsi please review

### DIFF
--- a/app/assets/sass/components/_filters.scss
+++ b/app/assets/sass/components/_filters.scss
@@ -1,10 +1,10 @@
 .filters {
 	@include rem('margin-bottom', $global-padding);
 
-	.col-sm-12 + .col-sm-12 {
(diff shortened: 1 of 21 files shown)


jeudy100:
@anujsi This MR is pretty much complete. I changed this to WIP because I am having a performace issue when I load all the user accounts. I get something like 7500+ users and trying to load them in the chosen select dropdown. Is this a normal amount of users or is it just because I have an internal account?

anujsi:
Remove console.log
some code is duplicated in filterByCreatorSelect.js and filterByOwnerSelect.js file so we can add in another file and try to reuse it.

jeudy100:
@anujsi I made a filterservice to keep the common code and I checked prod for the users, as well. I was only getting 87 users on prod, so I am going to take that as an indicator that there will not be thousands of users.

PR #958, opened 2019-06-25
PCM-7065: added recently selected products when creating a new case

jeudy100:
@anujsi please review

P.S. I asked akshay and pranjal if there should be a client side limit or if the server automatically limits the list of recently selected products. I will need to add a limit if the server doesn't automatically send a limited list.

### DIFF
--- a/app/cases/controllers/accountSelect.js
+++ b/app/cases/controllers/accountSelect.js
@@ -37,7 +37,7 @@ export default class AccountSelect {
                             ProductsService.clear();
                             CaseService.clearProdVersionFromLS();
                         }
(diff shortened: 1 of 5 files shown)


anujsi:
@jeudy100 did you add API `hydrajs.products.getProductVersionsV2` in the hydrajs librabry?

jeudy100:
@anujsi I'm pretty sure that I did, is that not the right products endpoint that I should be using?

vrathee:
@jeudy100 @asonawan Wanted to get some wires for this.. Was it done?

PR #963, opened 2019-06-28
PCM-2764: fixed not copying line breaks in discussion comments

jeudy100:
@anujsi please review


vrathee:
@jeudy100 We might also want to do it for other text area fields like description, case summary etc..

PR #986, opened 2019-07-17
PCM-7968: can add multiple files with ctrl+click

jeudy100:


app/cases/controllers/attachLocalFile.js
```diff
@@ -49,44 +49,45 @@ export default class AttachLocalFile {
         $scope.selectFile = function () {
             const minSize = 0;
```
rh-kgarg: @jeudy100 If i select 3 files at first and then without uploading them i try to select 3 more files from the file explorer, will it override the previously selected files?

jeudy100: @rh-kgarg It doesn't, it calls AttachmentsService.addNewAttachment for each file which just adds that to the end of the attachments array.

app/cases/views/attachLocalFile.jade
```diff
@@ -11,7 +11,7 @@
     ng-click='getFile()',
     ng-disabled='CaseService.addingComment',
     translate='',
-    ng-hide='AttachmentsService.updatedAttachments.length > 0') Attach File
+    ng-hide='AttachmentsService.updatedAttachments.length > 0') Attach file(s)
```
jeudy100: @rh-kgarg I talked to @KrystalMoore16 and we decided to switch the "Attach file" button string to "Attach file(s)" to show that the user can now upload multiple files.

anujsi:
Reviewed.

PR #990, opened 2019-07-25
PCM-7729: changed css modal text and made modal not close

jeudy100:


anujsi:
@jeudy100 You have changes text also in the pre element.
did you confirm it because I do not see any text changes in the linked Jira

jeudy100:
@anujsi Jason Willeford commented on the and wanted the error message to be changed to what it is now.

PR #994, opened 2019-08-06
PCM-6838: fixed action plan default size being too small

jeudy100:


rh-kgarg:
@asonawan Could you please review the css changes?

jeudy100:
@rh-kgarg I just noticed that I didn't push some changes I made that had a bunch of functionality that I took out because we decided to make the text box exactly like the public/private case comments input box under "Case Discussion" section.

anujsi:
Reviewed.

PR #1002, opened 2019-08-22
use correct condition to check if case is escalated or not

rh-kgarg:
@anujsi @vrathee @jeudy100 Please review.


anujsi:
Reviewed.

PR #1008, opened 2019-08-27
PCM-1416: added style changes to text line-height and link font size

jeudy100:
@rh-kgarg please review

### DIFF
--- a/app/assets/sass/components/_case-discussion.scss
+++ b/app/assets/sass/components/_case-discussion.scss
@@ -132,7 +132,7 @@
 			text-transform: uppercase;
 			@include font-size(13px);
 			font-size: 13px;
-			line-height: 1;
+			line-height: 1.5;
 			font-weight: 700;
 		}
 	}
@@ -142,7 +142,7 @@
 		font-weight: 700;
 		@include font-size(13px);
 		font-size: 13px;
-		line-height: 1;
+		line-height: 1.5;
 		@include rem('margin-bottom', 5px);
 	}
 
--- a/app/assets/sass/elements/_links.scss
+++ b/app/assets/sass/elements/_links.scss
@@ -6,6 +6,12 @@
 		color: darken($link-color, 5);
 	}
 }
+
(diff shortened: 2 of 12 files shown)


anujsi:
@asonawan Could you please review this PR because CSS related changes.

rh-kgarg:
@asonawan Could you please review this?
even though @jeudy100 has picked up the css changes from the jira's acceptance criteria but still we are not sure what might be the impact of such a change.

app/assets/sass/elements/_links.scss
```diff
@@ -6,6 +6,12 @@
 		color: darken($link-color, 5);
 	}
 }
+
+// breadcrumbs:last-child is needed because of the last bread crumb in the portal header
+a, #breadcrumbs:last-child {
```
asonawan: @jeudy100 is there any reason we are targeting anchor tag globally and changing the font size?

app/assets/sass/components/_case-discussion.scss
```diff
@@ -142,7 +142,7 @@
 		font-weight: 700;
 		@include font-size(13px);
 		font-size: 13px;
```
asonawan: I think this line is not required because we are already using '@include font-size(13px)'

app/assets/sass/components/_case-discussion.scss
```diff
@@ -132,7 +132,7 @@
 			text-transform: uppercase;
 			@include font-size(13px);
 			font-size: 13px;
```
asonawan: I think this line is not required because we are already using '@include font-size(13px)'

app/assets/sass/elements/_links.scss
jeudy100: in the jira it says that all links are too small and should be bigger, so I increased all link sizes to be 16px. The 16px is specified in the jira as well.

app/assets/sass/components/_case-discussion.scss
jeudy100: yea that is not needed

PR #1015, opened 2019-09-09
PCM-8510: fixed translations not showing up

jeudy100:


jeudy100:
@rh-kgarg @anugsi I have fixed the translation problem while also keeping the ui/ux the same. The only problem that I have now is with the chinese translations and that the solution engine text is translated into chinese while every other language has "Solution Engine" in english. I need to get in touch with the translation team so I can figure out which symbols mean solution engine so I can replace it with an a tag in the .po files.

app/cases/controllers/list.js
```diff
@@ -34,7 +35,7 @@ export default class List {
             try {
                 const accountNumber = SearchCaseService.searchParameters.accountNumber || securityService.loginStatus.authedUser.account_number;
                 const query = `case_accountNumber:${accountNumber}`;
-                const response = await strataService.cases.advancedSearch(query, null, 0, 1000, 'csv', solrCaseFields, null, null, null, null, null, 'lastModifiedDate');
+                const response = await strataService.cases.advancedSearch(query, null, 0, 10000, 'csv', solrCaseFields);
```
rh-kgarg: reason for this revert?

jeudy100: @rh-kgarg I think I might have branched off of a branch that wasn't updated to master, so that shouldn't have been changed. Already pushed the change to fix it.

rh-kgarg:
Reviewed. Please also create a QA tag.

PR #1028, opened 2019-10-09
PCM-8633 - In support cases, set both comments and input text box to fixed-width font

asonawan:
Recently we have changed our base font family to 'Red Hat Text' from 'Overpass' while doing that we have also changed the default font family of pre and form elements to  'Red Hat Text' from 'monospace' font hence the alignment of texts (specifically logs) were broken.


anujsi:
Reviewed.

PR #1033, opened 2019-11-13
show cluster id on details page if present in response irrespective of product selected

rh-kgarg:
@vrathee @anujsi Please review.
We will show cluster id on details page if it is present in case details response even if selected product is not Openshift Online.But if product is not openshift online and no cluster id is set we will not show it. This is how SFDC is doing it and Eric Rich suggested doing it the same way in PCM.


vrathee:
Reviewed!

PR #1045, opened 2020-01-17
minor issue for showing html/xml in  markdown .

anujsi:
@vrathee Please review.


vrathee:
Apart from renaming the function rest looks good!

